PR #1, opened 2017-02-11
Update how now playing notification is shown.

Tapchicoma:
Move all logic to the DownloadServiceImpl as it is not a responsibility of SubsonicTabActivity to do it.
Also use NotificationCompat.Builder to build notification as now it is a standard way to do it.


ogarcia:
LGTM (sorry for delay, hard weekend)

PR #56, opened 2017-10-17
Add get cover art

Tapchicoma:
#9 Add new getCoverArt method and use it in RESTMusicService


Tapchicoma:
Jacoco can't handle inlined functions

PR #96, opened 2017-12-16
Update with store badges instead of links.

Tapchicoma:

Thanks to @poussinou


Poussinou:
Thanks :)

PR #132, opened 2018-01-13
Fix api error parses always in generic error.

### LINKED ISSUE #131: Subsonic API doesn't parse error well
## Problem description

I've noticed that Subsonic api implementation always parses api into generic with code 0.

### Steps to reproduce

- Use wrong username/password
- Subsonic api will produce `SubsonicError.GENERIC` while it should be `SubsonicError.WRONG_USERNAME_OR_PASSWORD`

## Additional notes

Required for #121

Tapchicoma:
Now it parses to right error representation. Also fix that right
exception for error is not thrown on api call.

Closes #131


ogarcia:
@Tapchicoma If you wish, when this PR is finished, we can merge develop into master and make a new 2.1.1 release to have this new error handling functionality. I think that this will prevent most of issues that actually are opening.

UPDATE: What do you think about add a new error message when the user have a bad (or self-signed) certificate and not check the _Allow self-signed HTTPS certificate_ option?

Tapchicoma:
@ogarcia I think it can wait for the normal release.

Though if you still want to release don't forget to decrease `targetSdk` to 22 version.

Tapchicoma:
About *self-signed certificates* error - it may be not a bad idea, but let's see how many users will open issue for this kind of problem.

PR #144, opened 2018-01-23
Added Travis CI automatic builds

ogarcia:
Now when adding a new tag, Travis CI automagically build a release signed apk and add it to assets.

This is very useful to have an apk in [the releases](https://github.com/ultrasonic/ultrasonic/releases) for use in the bug support.


ogarcia:
@Tapchicoma in `.travis.yml` file there the following two variables:
```
- ANDROID_API=27
- ANDROID_BUILD_TOOLS=26.0.2
```
We must remember set it every time that we change the same values in `dependencies.gradle`.

ogarcia:
Taking advantage of the fact that you have not merged this PR, I added the Transifex suport. Now thanks to Travis CI, in every push of develop or master the changes in translation files are pushed automagically to Transifex too.

PR #150, opened 2018-01-24
 Added Transifex automation for localizations

ogarcia:
Using CircleCI.

The changes in language files are caused by purals support in Transifex that is standard and is fixed for every language. In this case we passed from `No songs` to `0 songs` that not is a problem.

This automation PUSH new changes in language files to Transifex for the community work on it. But the pull must be done "by hand" using `tx pull` command.

Take note of quotes in "by hand", because is simply execute a command that we can do before every release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,4 +66,9 @@ workflows:
       - push_translations:
           requires:
             - build
+          filters:
+            branches:
+              only:
+                - develop
+                - master
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,4 +38,32 @@ jobs:
           destination: reports
       - store_artifacts:
           path: build/reports/jacoco/jacocoFullReport/
+  push_translations:
+    docker:
+      - image: circleci/python:3.6
+    working_directory: ~/ultrasonic
+    steps:
+      - checkout
+      - run:
+          name: install transifex client
+          command: |
+            python -m venv ~/venv
+            . ~/venv/bin/activate
+            pip install transifex-client
+      - run:
+          name: configure transifex client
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -38,4 +38,32 @@ jobs:
           destination: reports
       - store_artifacts:
           path: build/reports/jacoco/jacocoFullReport/
+  push_translations:
```
Tapchicoma, line 41: Can you limit this job only to `master` and `develop` branches?

There is no point to do it for PR branches.

.circleci/config.yml
```diff
@@ -38,4 +38,32 @@ jobs:
...
+            . ~/venv/bin/activate
+            tx push -st
+workflows:
+  version: 2
+  build_and_push_translations:
+    jobs:
+      - build
+      - push_translations:
+          requires:
+            - build
```
Tapchicoma, line 68: Not sure that `build` here is right one job 🤔 

I think better to use a separate job that builds release type, but for now it is ok.

Tapchicoma (changes requested):

.circleci/config.yml
ogarcia: Done.

But we have disabled the option _Build forked pull requests_, so we don't worry about third parts PRs.

.circleci/config.yml
ogarcia: I think that better workflow is sequential like this:
1. Do tests.
2. Build.
3. Push translations (because the project build well).
4. If is a tag `/[0-9]+(\.[0-9]+)*/` sign build and publish it in release.

.circleci/config.yml
Tapchicoma: Still, for a example, I can add some translation in my PR and then remove it, so better to have it enabled just for `master` and `develop` branches.

Tapchicoma (approved):
LGTM

PR #152, opened 2018-01-25
Added Circle CI automatic builds

ogarcia:
First merge #150 to see real changes

Now when adding a new tag, Travis CI automagically build a release signed apk and add it to assets.

For now this use two new jobs, the first `generate_signed_apk` runs a `./gradlew build assembleRelease` and signs the APK with `ultrasonic-keystore` that is encoded with a random password (stored in CI variables).

The second job `publish_signed_apk` uses a small tool called [ghr](github.com/tcnksm/ghr) to publish the signed APK in the tag.

I think that in next iterations we can do it better, but for now it works well.


.circleci/config.yml
```diff
@@ -38,4 +38,93 @@ jobs:
...
+          name: push changes in translation files
+          command: |
+            . ~/venv/bin/activate
+            tx push -st
+  generate_signed_apk:
+    docker:
+      - image: circleci/android:api-27-alpha
+    working_directory: ~/ultrasonic
+    envoronment:
+      JVM_OPTS: -Xmx3200m
```
Tapchicoma: Better to change to:
```
JAVA_TOOL_OPTIONS: "-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap"
```

Check this article: https://circleci.com/blog/how-to-handle-java-oom-errors/

.circleci/config.yml
```diff
@@ -38,4 +38,93 @@ jobs:
...
+            for apk in *.apk; do
+              jarsigner -verbose -sigalg SHA1withRSA -digestalg SHA1 -keystore ~/ultrasonic/ultrasonic-keystore -storepass ${ULTRASONIC_KEYSTORE_STOREPASS} -keypass ${ULTRASONIC_KEYSTORE_KEYPASS} ${apk} ultrasonic
+              jarsigner -verify ${apk}
+              ${ANDROID_HOME}/build-tools/27.0.0/zipalign -v 4 ${apk} /tmp/ultrasonic-release/${apk/-release/}
+            done
+      - persist_to_workspace:
+          root: /tmp/ultrasonic-release
+          paths:
+            - ultrasonic-*.apk
+  publish_signed_apk:
```
Tapchicoma: Can you rename to `publish_github_release`? Because initially I though that it publishes to play store.

.circleci/config.yml
```diff
@@ -38,4 +38,93 @@ jobs:
...
+            - build
+          filters:
+            branches:
+              only:
+                - develop
+                - master
+      - generate_signed_apk:
+          filters:
+            tags:
+              only: /^[0-9]+(\.[0-9]+)*/
```
Tapchicoma, line 119: If I (as example) will push by mistake some tag to my PR - will it also make a new release? Maybe worth to limit again only to `develop` and `master` tags.

Tapchicoma (changes requested):
Nice job!

.circleci/config.yml
```diff
@@ -38,4 +38,93 @@ jobs:
...
+      - run:
+          name: build release apk
+          command: ./gradlew build assembleRelease
+      - run:
+          name: sign release apk
+          command: |
+            mkdir -p /tmp/ultrasonic-release
+            cd ultrasonic/build/outputs/apk/release/
+            for apk in *.apk; do
+              jarsigner -verbose -sigalg SHA1withRSA -digestalg SHA1 -keystore ~/ultrasonic/ultrasonic-keystore -storepass ${ULTRASONIC_KEYSTORE_STOREPASS} -keypass ${ULTRASONIC_KEYSTORE_KEYPASS} ${apk} ultrasonic
```
Tapchicoma, line 83: Maybe modify `ultrasonic\build.gradle` to following:
```gradle
signingConfigs {
        release {
            storeFile file('ultrasonic.keystore')
            storePassword System.getenv('ULTRASONIC_KEYSTORE_STOREPASS')
            keyAlias System.getenv('KEY_ALIAS')
            keyPassword System.getenv('KEY_ALIAS_PASSWORD')
        }
}
```
So on building release it will be automatically signed.

Usually it is preferable to use gradle signing job then `jarsigner` directly.

Tapchicoma (changes requested):

.circleci/config.yml
ogarcia: A tag is like a branch in git, you can select what branches or tags starts the CI process but not are dependent one of other. If you set a filter with `master` and `develop` and tags then it works in any push to `master` or `develop` or tag.

Anyway in this case if you push a tag, CI only works if the tag is like a version `x.x..`.

.circleci/config.yml
ogarcia: Yes, but the Google Play key not is the same and if we use `signingConfigs` in `build.gradle` then I must change values every time that we want to publish a new release.

.circleci/config.yml
ogarcia: Done!

.circleci/config.yml
ogarcia: Done! Better `publish_github_signed_apk` that is more descriptive :wink:

.circleci/config.yml
Tapchicoma: We can introduce flavors for Google Play, Github, F-droid that can contain it own singing configs (except F-droid). But current approach is fine for me for now.

Tapchicoma (approved):

PR #154, opened 2018-01-29
Fix F-Droid

ogarcia:
As we can see in https://gitlab.com/fdroid/fdroiddata/issues/1107 seems that use autoversioning breaks F-Droid builds.

This PR reverts changes introduced in https://github.com/ultrasonic/ultrasonic/commit/c84482729d216d2b2f1c104f7b89de68806e17e5 and sets `versionCode` and `versionName` in `dependencies.gradle`.

And with this seems that builded apk name is `ultrasonic-release.apk` instead of `ultrasonic-release-VERSION.apk`. If finaly we merge this I must change the Circle CI autobuild to use the new name.


Tapchicoma:
I see, sad limitation of F-Droid. At least that can try to do it more intelligent, like parse it from final merged `AndroidManifest.xml` or from apk itself.

Tapchicoma:
Just for history - issue in Gitlab build server: https://gitlab.com/fdroid/fdroidserver/issues/388

Issue #98, opened 2017-12-19
Reimplement custom cache folder setting


Tapchicoma:
## Problem description

To use custom cache folder user needs to type path on the device (for example `/sdcard/ultrasonic`) that requires advanced Android knowledge and not user-friendly.

App need to introduce another way to set cache folder.

*Note:* using external storage will require granting runtime permission.

## Possible Improvements:
- Select cache folder for each server individually
- Provide dialog or separate screen to select desired cache folder (user may want to create a new one)
- Call external app to do this selection.
- Request permission before selecting the folder.

ogarcia:
@Tapchicoma There is a lot of people that say in the comments that they want use an external storage (sdcard) for Ultrasonic cache. Is true that the actual method is not user friendly.

As you say in #79 is a good idea remove dangerous permissions and request it only if the user wants use non-default cache location.

By default the app may have a simple selector for use internal storage or sdcard and an advanced setting for define "by hand" (or using external browser app) the cache path.

Another improvement may be the possibility of delete a full album or artist from cache.

Even redefine the controllers _pin_, _unpin_ and _download_ that not is clear. I think that _pin_ implies download and _unpin_ implies delete but, at this moment, the app not works like that.

Whisprin:
This is how it looks currently in [DSub](https://github.com/daneren2005/Subsonic): [Cache location](https://user-images.githubusercontent.com/3153136/34182769-e89d1cea-e517-11e7-9475-503ae6dc4a09.png)
But adjusting this path, which is writeable by the app without requesting special permissions, to match Ultrasonic brings up an error message and resets the path to default. I'm running Android 8.0.

Additionally `/sdcard/ultrasonic` works but is on _internal_ memory.

An up to date implementation to request rw access to the _external_ memory can be found in the [Amaze file manager](https://github.com/TeamAmaze/AmazeFileManager).

ogarcia:
The DSub idea is good, but I think that instead of show location by default, have a selector with 3 options, Internal, External and _Advanced manual selection_. If you enter in last option then show the path and give the possibility of change it.

Tapchicoma:
To make it easier to update, let's go with selector that have 2 entries:
- Internal storage (app private internal storage)
- External storage (app private external storage)

Later if there will be request from users app can introduce 3rd option "Custom location"

PR #159, opened 2018-02-07
Offline requests support

Tapchicoma:
#98: Initial changes required for updating app caching.

With this change app will put each api response into the http client cache, even if server prohibits it in headers.

When device is online - app will make common network call, but when device is offline and response is in the cache - it will return response from cache. Response will be valid in cache for offline usage for around 3 months.

My intention is to move app from current custom file based caching and later think about additional DB layer for entities.

As this change is pretty controversial, I requested @ogarcia review.


ogarcia:
The main problem about this is that you are modifying the response headers of server. This can cause unexpected or wrong behaviors, or even problems when you modify your library in the server.

I think that the cache must be restricted only to music files. The app have an offline mode that allows navigate the cached music. But if we add this change the app will show entire cached info (entire library in most cases) and this whould be confusing for non advanced users, that they will see music that cannot access.

Another point is what you say about database. This is a great idea, build a database based on downloaded music that could provide better offline navigation.

But any changes about offline mode must be based only on music files. If a music file is in cache folder then appears in offline mode, but if the music file not is present no information should appear in any case because it will be confusing to user.

@Tapchicoma what do you think?

Tapchicoma:
@ogarcia As I said it is only initial step to better support offline mode.

> The main problem about this is that you are modifying the response headers of server. This can cause unexpected or wrong behaviors, or even problems when you modify your library in the server.

I may update `Interceptor` to be more intelligent and don't remove `max-age=` and `max-stale=` fields from header, but still allowing to cache responses. Current logic is to allow caching, but use them only when device is offline, otherwise make a new network call.

> I think that the cache must be restricted only to music files. The app have an offline mode that allows navigate the cached music. But if we add this change the app will show entire cached info (entire library in most cases) and this whould be confusing for non advanced users, that they will see music that cannot access.

The more user uses the app - the more responses will be in the cache and less confusing it will be for the user. Main idea for this change is to allow for app to continue working in short-term internet disconnection (like riding on the DB train and having shitty internet connection 😄 )

Later introduction of DB layer will allow to more proper implement real "offline" support.

ogarcia:
> I may update `Interceptor` to be more intelligent and don't remove `max-age=` and `max-stale=` fields from header, but still allowing to cache responses. Current logic is to allow caching, but use them only when device is offline, otherwise make a new network call.

IMHO, the app never would rewrite the server headers. Use cache is a great idea, but remove `Cache-Control` and `Pragma` and rewrite `Cache-Control`, may produce unexpected behabiours. I think that is better show a "Sorry, your conection blah blah..." message that have problems with lists that not update. If the server (or the server admin), want no use cache for whatever reason, we should not force it.

Tapchicoma:
@ogarcia The general problem that there are a lot of Subsonic API implementations and it is hard to convince them to introduce at least **some** caching support. And again *I am pointing out that this only affects the time when device is offline*, so when device is online and made this call - it should rewrite it in cache with new one, though I see possible problem with calls that modify some state on the server side.

What about if I add better header overwriting strategy and enable this caching support only for API get* type methods (that just get info from server, but not update/modify anything - this calls will fail in offline mode)?

ogarcia:
> The general problem that there are a lot of Subsonic API implementations and it is hard to convince them to introduce at least **some** caching support. And again *I am pointing out that this only affects the time when device is offline*, so when device is online and made this call - it should rewrite it in cache with new one, though I see possible problem with calls that modify some state on the server side.

The main problem is if any sysadmin make his/her own header configuration for any reason and we rewrite them. Take note that I can make a setup like a subsonic server behind a nginx reverse proxy, and configure in nginx a forced no cache control. May we, in these cases, overwrite the user configuration? I think we should no.

> What about if I add better header overwriting strategy and enable this caching support only for API get* type methods (that just get info from server, but not update/modify anything - this calls will fail in offline mode)?

The problem is the same, I think that if the server admin have own configuration without cache support or forcing no cache we not overwrite it.

Anyway if you work with cache in update or modify methods is even worse, because when you go online the changes will be lost.

IMHO. I would add the cache support like in a browser and leave to users the choice te enable it or not in his/her own servers setup.

ogarcia:
And testing the code I always receive this error in first connection test (after configure new server):
```
02-10 11:02:32.072 15573-15573/org.moire.ultrasonic W/ServerSettingsFragment: java.lang.RuntimeException: Failed to create requests cache
                                                                              java.lang.RuntimeException: Failed to create requests cache
                                                                                  at org.moire.ultrasonic.service.MusicServiceFactory.getCacheDirectory(MusicServiceFactory.java:125)
                                                                                  at org.moire.ultrasonic.service.MusicServiceFactory.createSubsonicApiClient(MusicServiceFactory.java:96)
                                                                                  at org.moire.ultrasonic.service.MusicServiceFactory.getMusicService(MusicServiceFactory.java:65)
                                                                                  at org.moire.ultrasonic.fragment.ServerSettingsFragment$1.doInBackground(ServerSettingsFragment.java:246)
                                                                                  at org.moire.ultrasonic.fragment.ServerSettingsFragment$1.doInBackground(ServerSettingsFragment.java:235)
                                                                                  at org.moire.ultrasonic.util.ModalBackgroundTask$3.run(ModalBackgroundTask.java:93)
```

Tapchicoma:
Ok, I see your point.

Closing this PR.

PR #163, opened 2018-02-10
New 2.2.2 release

Tapchicoma:
Contains bugfix for stray ">" and fixes for better cache handling.


ogarcia:
Still failing the file name in downloads. If you download any song it not have extension. Only `song_name.complete.`. Then if you go to offline mode, no songs appear because not have valid extension :cry:

ogarcia:
Make some debug I see that is picking in `FileUtil.java` the `TranscodedSuffix`, but variable is empty. If the song is not transcoded I think that variable must be null as I undestand by this piece of code.

```
		if (song.getTranscodedSuffix() != null)
		{
			fileName.append(song.getTranscodedSuffix());
		}
		else
		{
			fileName.append(song.getSuffix());
		}
```

Or change this if to some like this `if (song.getTranscodedSuffix() != null && !song.getTranscodedSuffix().isEmpty())`. I prefer that you take a look.

Tapchicoma:
@ogarcia checked - now app uses right suffix for fully download song.

ogarcia (approved):
LGTM

PR #173, opened 2018-03-10
Persistent cache

Tapchicoma:
Part of #98 - adds persistent cache that can store/load domain entities.


Tapchicoma:
Blocked by: https://github.com/twitter/Serial/issues/31

domain/src/main/kotlin/org/moire/ultrasonic/domain/MusicFolder.kt
```diff
@@ -6,4 +6,6 @@ package org.moire.ultrasonic.domain
 data class MusicFolder(
         val id: String,
         val name: String
-)
+) {
+    companion object
```
Tapchicoma: Not needed.

PR #178, opened 2018-03-18
New 2.3.0 release

Tapchicoma:
Not so much visible changes, but worth to make a new release.


Tapchicoma:
BTW: as I enabled minify for release build - better to test release instead of debug.

ogarcia (commented):
LGTM

PR #182, opened 2018-03-21
New 2.3.1 release

Tapchicoma:
Checked locally, for me release build is now working.


ogarcia (approved):
LGTM.

I'm going to publish it in BETA release in Google Play Store. I think that is better for easy rollback if we found any bug.

I think that if in some days (One week?) no bugs repoted then move it to stable release.

Issue #238, opened 2018-11-16
Ultrasonic can't connet to Subsonic 6.1.4


djnaoki-mrn:
## Problem description

Since the last update of subsonic (6.1.4) ultrasonic can't connect to the serveur. i tryed with others app and it's work fine.

## System information

 * **Ultrasonic version**: *2.5.0*
 * **Android version**: *8.1.0*
 * **Device info**: *OnePlus 5T*

srayneau:
Same problème after update subsonic server to 6.1.4
Message in french "Connection échouée une erreur réseau est survenue,veuillez vérifier l'adresse du serveur ou réessayer plus tard."
-> try reconfigure new server -> ko
-> try to unistall / install the application again -> ko
-> Work with subsonic application.

djnaoki-mrn:
hi, for add more information, when i want to connect the app to my server there are no reaching time, the app say directly that there are an issue with the network... if i shutdown the server, the app take a little time to say that there are an issue with the network.
it's as if the server directly reject the connection.
is it possible that the app use an old way of the api for the connection ? 

thanks

this is a screenshot of the message : https://ibb.co/hfwxZf

djnaoki-mrn:
just for information ultrasonic also bug with the official demo server : http://demo.subsonic.org
usr: guest
pwd: guest

ogarcia:
@Tapchicoma is a problem with API version. Latest version of Subsonic reports API version `1.16.1`.

Here is the error:
```
11-19 09:11:49.621 2026-2026/org.moire.ultrasonic.debug W/ServerSettingsFragment: com.fasterxml.jackson.databind.JsonMappingException: Unknown api version 1.16.1 (through reference chain: org.moire.ultrasonic.api.subsonic.response.SubsonicResponse["version"])
    com.fasterxml.jackson.databind.JsonMappingException: Unknown api version 1.16.1 (through reference chain: org.moire.ultrasonic.api.subsonic.response.SubsonicResponse["version"])
```

The Subsonic changelog says that new point added:
* REST: Added artistImageUrl to Artist and ArtistID3

The changes in was in Artist:
```xml
<xs:complexType name="Artist">
 <xs:attribute name="id" type="xs:string" use="required"/>
 <xs:attribute name="name" type="xs:string" use="required"/>
 <xs:attribute name="artistImageUrl" type="xs:string" use="optional"/>
 <!-- Added in 1.16.1 -->
[...]

PR #239, opened 2018-11-22
Fast and dirty fix to Subsonic 6.1.4

ogarcia:
Only use the new API level as a lower level. Not is elegant but works as workaround of #238 

@Tapchicoma please give your geen light to merge it to master


ogarcia:
@Tapchicoma i go to merge this, is a very small change

Tapchicoma:
@ogarcia sorry, was not around my computer. Seems in `1.16.1` new two optional fields in Artist model was added, so fix is pretty safe. In general, I think app should better behave in future to such api changes.

PR #261, opened 2020-01-12
Implemented five star rating feature for songs

nitehu:
This was requested in #260, also I really needed this feature.
I'm not an android developer, but I hope the code can be used.
Please let me know if something should be changed.

Issue #260, opened 2019-12-04
FR allow setting ratings for songs, album, artist


lachlan-00:
I only just noticed this but you can only flag/star items in the UI.
Not being able to set a rating is a pretty big missing feature.

PR #263, opened 2020-01-14
Implemented five star rating feature for songs

nitehu:
This was requested in #260, also I really needed this feature.
I'm not an android developer, but I hope the code can be used.
Please let me know if something should be changed.
(This pull request is the duplicate of #261, but now the commit is signed)


ogarcia (approved):
LGTM

palto42:
I would really appreciate if this feature is added to Ultrasonic. I'm using it in my own build since 2 month and it works without issues.

Issue #79, opened 2017-11-19
Support runtime permissions


Tapchicoma:
In order to update api level app should support runtime permissions.

Currently app uses following `dangerous` permissions:
- android.permission.WRITE_EXTERNAL_STORAGE
- android.permission.READ_EXTERNAL_STORAGE

Tapchicoma:
This permission is only required when user defines non-default cache location. 

The problem is that current setting for this requires typing full path to desired cache folder (for example `/sdcard/ultrasonic`) and it is not very usable. 

I am considering removing this setting, removing permission and just use device app private location on external storage for cache.

@ogarcia will such approach be ok?

Tapchicoma:
Created separate ticket for this setting: #98

ogarcia:
As I say in #98 IMHO I think that the first approach must be a simple selector that permits choose between internal storage and sdcard

ogarcia:
@Tapchicoma I add `help wanted` label in case someone wants to help with a PR :wink:

Tapchicoma:
@ogarcia sorry, don't really have time for addressing main blocker for now to advance app further - how to migrate from current caching approach to new one that will not require storage permission (kind of a lot changes) :disappointed_relieved: 

This is one of the "need-to-do" for bumping app target version to latest one and re-enable publishing it to play store.

ogarcia:
@Tapchicoma no problem, this is a voluntary project and we always have periods without free time. If someone can help us perfect, if not, then it will be addressed in the future. :blush:

PR #278, opened 2020-06-02
Implemented runtime permission support

nitehu:
This is my take on #79 .

It contains the following:
- Migrated to Androidx (libraries seem to need this)
- Added MultiDex because with Androidx there were too many functions for a single dex
- Added Dexter library, which seems to be a well-maintained permission handling library for Android
- Implemented runtime permission handling

The added permission handling logic:
- is backward compatible, before android M it works as usual
- only affects the music cache directory, all other files are stored in the app's external directory and require no additional permissions
- when the music cache directory is set to its default value, Ultrasonic doesn't need additional permissions and won't ask for them
- when the music cache directory is changed, or anytime Ultrasonic can't access it, it will request permissions
- while the permission request is in progress, Ultrasonic will use the default directory for cache
- if permission is given, Ultrasonic will use the new cache directory, otherwise it will reset the cache to the default location

Executed tests:
- Emulators: Galaxy Nexus API17, Pixel 2 API 27
- Phones: Samsung Nexus S with Android 4.4, Honor 8 with Android 7.0, Galaxy Note 9 with Andoid 10

Known problems:
- It seems the Ultrasonic directory must be changed (to ...org.moire.ultrasonic/*files*) because without permissions only this directory can be accessed. The old files can be cleaned up from android settings. For new installations and reinstalls this is not a problem.
- If the user revokes the permission, Ultrasonic will use the default cache location temporarily while it asks for permissions again. I'm not sure what else it could do...
- (The new functions aren't using Kotlin. It seems the UI related things are still in Java, and I didn't want to force it (also I'm not good with Kotlin))
- (I'm not sure how can I write unit tests for this)

Please review and let's talk about this if something needs to be changed.


ogarcia:
Hi @nitehu 

First for all, many thanks for work!!! I'll review it along this week :wink:

build.gradle
```diff
@@ -13,7 +13,7 @@ buildscript {
         maven { url "https://plugins.gradle.org/m2/" }
     }
     dependencies {
-        classpath gradlePlugins.androidTools
+        classpath "com.android.tools.build:gradle:3.6.3"
```
Tapchicoma: Why have you changed your version here and not in `dependencies.gradle`? :thinking:

ultrasonic/build.gradle
```diff
@@ -13,6 +13,7 @@ android {
 
         minSdkVersion versions.minSdk
         targetSdkVersion versions.targetSdk
+        multiDexEnabled true
```
Tapchicoma: Is it really needed for the app? App should be not so big to require multidex :thinking:

build.gradle
nitehu: Oh sorry that was Android Studio and I didn't notice it.
Also I forgot to check in dependencies.gradle, I fixed that too.

ultrasonic/build.gradle
nitehu: After migrating to Androidx, the build was throwing errors because of more functions than 65535, and suggested to use multidex.
I'll try to test it again.

Tapchicoma: Interesting :thinking: 

Would be interesting to see what dependency brings so many methods count. This project does not have many dependencies.

ogarcia: @Tapchicoma I'm doing some updates to use gradle '6.5.0' but breaks all tests, seems that is a minor change but I cannot find the key

dependencies.gradle
```diff
@@ -1,18 +1,18 @@
...
         gradle               : '4.6',
 
-        androidTools         : "3.2.0",
+        androidTools         : "3.6.3",
         ktlint               : "0.28.0",
         ktlintGradle         : "5.1.0",
         detekt               : "1.0.0.RC6-4",
         jacoco               : "0.8.2",
 
         androidSupport       : "22.2.1",
```
ogarcia, line 13: You can change this to:
```
androidSupport            : "1.0.0",
androidSupportDesign      : "1.0.0",
androidSupportAnnotations : "22.2.1",
```
And use this vars in `ext.androidSupport`

dependencies.gradle
```diff
@@ -41,9 +41,9 @@ ext.gradlePlugins = [
 ]
 
 ext.androidSupport = [
-        support     : "com.android.support:support-v4:$versions.androidSupport",
-        design      : "com.android.support:design:$versions.androidSupport",
-        annotations : "com.android.support:support-annotations:$versions.androidSupport"
+        support    : 'androidx.legacy:legacy-support-v4:1.0.0',
```
ogarcia: Replace with:
```
support     : "androidx.legacy:legacy-support-v4:$versions.androidSupport",
```

dependencies.gradle
```diff
@@ -41,9 +41,9 @@ ext.gradlePlugins = [
...
+        design     : 'com.google.android.material:material:1.0.0',
```
ogarcia: Replace with:
```
design      : "com.google.android.material:material:$versions.androidSupportDesign",
```

ultrasonic/build.gradle
nitehu: Checked the number of functions with dexcount, these namespaces contain the most functions:

```
methods  fields   package/class name
5533     853      android
15440    11354    androidx
11282    3220     com.fasterxml
3159     3668     com.google.android.material
20984    4727     kotlin
4606     5833     org.moire.ultrasonic
```

Overall Androidx, Kotlin and FasterXml are responsible for most of the 69643 total functions.

Tapchicoma: probably you could enable multidex only for debug builds. I will expect proguard will cut out most of this methods for release builds.

Probably app minimal supported sdk should be bumped to 21 to get rid of multidex at all.

@ogarcia old version of AGP could also be a problem when updating to newer Gradle.

nitehu: Yes, minSdk 21 would solve this entirely, but I didn't want to bump it up when it isn't absolutely necessary. Google says that ~94% of the devices are above or on Sdk 21.
What do you guys say, leave it (just change to debug only) or bump it?

ogarcia: I prefer maintain minSdk in 14 until we have a blocked feature. Not much users with Android 4, but still we have some.

ogarcia: @Tapchicoma If I update ktlint, ktlintGradle and jacoco can upgrade gradle to 6.5.0 and buil app without problems, but some test fail to execute

dependencies.gradle
nitehu: Ah OK, I see now, thank you!
Moved all dependencies to dependencies.gradle, with separated versions.

dependencies.gradle
nitehu: Fixed.

dependencies.gradle
nitehu: Fixed.

ultrasonic/build.gradle
nitehu: Moved multidex to debug.

ogarcia:
@Tapchicoma do you know why don't pass the `ktlintCheck` tests?

nitehu:
I think I've managed to solve the ktLintCheck problem, now the assembleDebug fails, I'm not sure why yet...
Oh I see, gradle.properties are set to "" for CI?

nitehu:
Wow. And the CI now works!

ogarcia:
Wow! Great work! 👏👏

@Tapchicoma I don't know if you want do something more. If not I'll merge this tomorrow

ogarcia:
Hi @nitehu I do several tests and found some issues.

- I cannot run this in API level 24, gives a `android.view.View$OnUnhandledKeyEventListener` and instant crash.
- If I set the targetSdk to 28 and APP crashes at launch because `getUltraSonicDirectory` in [FileUtil][1] returns `null`
- I do some changes to fix the _ForegroundService_, see [api28 branch][2] ([this commit][3]) to avoid repeat the work :wink:.

[1]: https://github.com/ultrasonic/ultrasonic/blob/af5ecae7f55634f58802d37b0dea92c93257c755/ultrasonic/src/main/java/org/moire/ultrasonic/util/FileUtil.java#L373
[2]: https://github.com/ultrasonic/ultrasonic/tree/api28
[3]: https://github.com/ultrasonic/ultrasonic/commit/af5ecae7f55634f58802d37b0dea92c93257c755

nitehu:
Yes I've tested it up to 23, as this pr only meant to fix the runtime permissions, and I planned to upgrade to 28 later :)
Let me know if I can help with anything.

Tapchicoma:
For me it looks good.

PR #279, opened 2020-06-17
Fixed context in FileUtil

nitehu:
Okay I see now what have you meant.
I was sooo happy that I didn't have to pass down the context so deep into the FileUtil. Well, it burned me :)
Also I've added usesCleartextTraffic="true" because the next thing that died was the connection to my server (that has no https configured).

With these changes it works now for me, but I didn't tested it extensively yet. I plan to go over Google's API28 checklist anyways.


ogarcia:
Yeahhhhhh :sunglasses:

I do merge this and #278 into develop to continue work more easily. The objetive is reach full API 28 support to close #79 and upload a new version to Google Play Store

Issue #283, opened 2020-06-22
Fix build release apk step in CI


ogarcia:
At this moment the command `./gradlew build assembleRelease` fails with several warnings and errors. We must fix it to build new releases to GitHub.

ogarcia:
Fixed!!! Thanks @nitehu :1st_place_medal:

PR #284, opened 2020-07-13
Refactored DownloadServiceImpl to use separate service for playback

### LINKED ISSUE #282: Refactor the DownloadServiceImpl to use a separate service for playback
As suggest @nitehu in #280 we need to refactor DownloadServiceImpl to use a separate service for playback to avoid the problems described in his pull request.

nitehu:
This fixes #282 (and possibly #283, on my machine at least).

This PR contains the following changes:
- The DownloadServiceImpl was refactored into many smaller classes. These classes are singletons injected with Koin. I tried to remove circular references between them but it isn't trivial and seems to be a longer project. The main offender for now is the JukeBox player. Maybe by further refactoring that the circular references can be removed. Not that they cause anything bad, just the code would be more clear.
- There is only one service now, the MediaPlayerService, which starts with the playback and stops when the player stops. It is a foreground service, and always displays a notification, as android likes it.

I'm using this version as my daily driver for a few weeks now, and I fixed every bug I encounterd during this time, there were some I created with the refactor and maybe some older too. This version seems to be reasonably stable but I'll continue monitoring and fixing if something arises.

As for the new classes, I know I'm bad at naming things, so let me know if I should rename anything for clarity, or if there are other changes to make this better.

### DIFF
--- a/ultrasonic/src/main/java/org/moire/ultrasonic/util/VideoPlayerType.java
+++ b/ultrasonic/src/main/java/org/moire/ultrasonic/util/VideoPlayerType.java
@@ -41,7 +41,7 @@ public enum VideoPlayerType
 				public void playVideo(final Activity activity, MusicDirectory.Entry entry) throws Exception
 				{
 
-					// Check if MX Player is installed.
+					// Check if MX LocalMediaPlayer is installed.
 					boolean installedAd = Util.isPackageInstalled(activity, PACKAGE_NAME_MX_AD);
 					boolean installedPro = Util.isPackageInstalled(activity, PACKAGE_NAME_MX_PRO);
 
--- a/ultrasonic/src/main/java/org/moire/ultrasonic/util/VideoPlayerType.java
+++ b/ultrasonic/src/main/java/org/moire/ultrasonic/util/VideoPlayerType.java
@@ -41,7 +41,7 @@ public enum VideoPlayerType
 				public void playVideo(final Activity activity, MusicDirectory.Entry entry) throws Exception
 				{
 
-					// Check if MX LocalMediaPlayer is installed.
+					// Check if MX Player is installed.
 					boolean installedAd = Util.isPackageInstalled(activity, PACKAGE_NAME_MX_AD);
 					boolean installedPro = Util.isPackageInstalled(activity, PACKAGE_NAME_MX_PRO);
 
--- a/ultrasonic/src/main/res/values-pt-rBR/strings.xml
+++ b/ultrasonic/src/main/res/values-pt-rBR/strings.xml
@@ -334,8 +334,8 @@
     <string name="util.bytes_format.megabyte">0.00 MB</string>
     <string name="util.no_time">-:--</string>
     <string name="util.zero_time">0:00</string>
-    <string name="video.get_mx_player_text">O Player MX não está instalado. Baixe da graça pela Play Store ou modifique as configurações de vídeo.</string>
-    <string name="video.get_mx_player_button">Baixar Player MX</string>
+    <string name="video.get_mx_player_text">O MX Player não está instalado. Baixe da graça pela Play Store ou modifique as configurações de vídeo.</string>
+    <string name="video.get_mx_player_button">Baixar MX Player</string>
     <string name="widget.initial_text">Toque para selecionar a música</string>
     <string name="widget.sdcard_busy">Cartão SD indisponível</string>
     <string name="widget.sdcard_missing">Sem cartão SD</string>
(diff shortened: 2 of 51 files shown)


ultrasonic/src/main/java/org/moire/ultrasonic/util/VideoPlayerType.java
```diff
@@ -41,7 +41,7 @@
 				public void playVideo(final Activity activity, MusicDirectory.Entry entry) throws Exception
 				{
 
-					// Check if MX Player is installed.
+					// Check if MX LocalMediaPlayer is installed.
```
ogarcia: Why this change? This checks if you have [MX Player][1] installed to play video.
[1]: https://play.google.com/store/apps/details?id=com.mxtech.videoplayer.ad

ultrasonic/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -334,7 +334,7 @@
     <string name="util.bytes_format.megabyte">0.00 MB</string>
     <string name="util.no_time">-:--</string>
     <string name="util.zero_time">0:00</string>
-    <string name="video.get_mx_player_text">O player MX não está instalado. Baixe da graça pela Play Store ou modifique as configurações de vídeo.</string>
+    <string name="video.get_mx_player_text">O localMediaPlayer MX não está instalado. Baixe da graça pela Play Store ou modifique as configurações de vídeo.</string>
```
ogarcia: This is the same: [MX Player][1]

[1]: https://play.google.com/store/apps/details?id=com.mxtech.videoplayer.ad

ogarcia:
Still have pending messages for review @nitehu

nitehu:
I'm sorry but may I ask what pending messages and where? Could you please help me find them?

ogarcia:
Here in GitHub PR, see code comments. Is about change MX Player to MX localMediaPlayer

nitehu:
Unfortunately I can't see any review comments! I also haven't received any notifications that there is a problem :(
But I've searched for "MX" and I've found what you mentioned, there are some unintended replaces. I'll fix this asap.

nitehu:
I'm a bit bugged by the review comments not showing up... Would you please check if this is what's happening?
https://stackoverflow.com/a/45336490/4207238

ogarcia (changes requested):
Checking the "Request changes"

ogarcia (approved):

ogarcia:
> I'm a bit bugged by the review comments not showing up... Would you please check if this is what's happening?
> https://stackoverflow.com/a/45336490/4207238

Yess. My fault.

ultrasonic/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -334,7 +334,7 @@
     <string name="util.bytes_format.megabyte">0.00 MB</string>
     <string name="util.no_time">-:--</string>
     <string name="util.zero_time">0:00</string>
-    <string name="video.get_mx_player_text">O localMediaPlayer MX não está instalado. Baixe da graça pela Play Store ou modifique as configurações de vídeo.</string>
+    <string name="video.get_mx_player_text">O Player MX não está instalado. Baixe da graça pela Play Store ou modifique as configurações de vídeo.</string>
```
ogarcia: This should be "MX Player" instead of "Player MX"

ultrasonic/src/main/res/values-pt/strings.xml
```diff
@@ -334,7 +334,7 @@
     <string name="util.bytes_format.megabyte">0.00 MB</string>
     <string name="util.no_time">&#8212;:&#8212;&#8212;</string>
     <string name="util.zero_time">0:00</string>
-    <string name="video.get_mx_player_text">O localMediaPlayer MX não está instalado. Descarregue da graça pela Play Store ou modifique as configurações de vídeo.</string>
+    <string name="video.get_mx_player_text">O Player MX não está instalado. Descarregue da graça pela Play Store ou modifique as configurações de vídeo.</string>
```
ogarcia: This should be "MX Player" instead of "Player MX"

ogarcia (changes requested):

ultrasonic/src/main/res/values-pt/strings.xml
nitehu: Yep those were strange for me too, but those were the original (maybe google translated) strings.
Anyway, they're fixed now.

ogarcia:
@nitehu sorry for the big delay. In holidays dates I have too much work

nitehu:
It's OK, I also don't have too much time.
I hope you can rest a bit too.

PR #290, opened 2020-08-18
Implemented file picker for cache folder settings

### LINKED ISSUE #98: Reimplement custom cache folder setting

nitehu:
This fixes #98 .

The cache folder setting was changed to a custom dialog. The dialog can
- request r/w permission when the user first wants to change the location of the cache;
- browse for a folder on the internal storage;
- set back the default internal application folder;
- create a new folder;
- KitKat+: recognize if there is an external SD card, and suggests the application folder on that - the user can't browse freely here, as Android forbids writing elsewhere
- before KitKat: SD cards can be browsed and selected as usual (this wasn't tested as I have no such an old phone nor emulator with sd card emulation)

As mentioned, after KitKat the only usable place on an SD card for Ultrasonic is its own application directory, which can be selected. To use other folders on the SD card it seems Ultrasonic must be updated to use the new Storage Access Framework. It seems this will be a requirement when upgrading to API29-30...

The current solution satisfies what is written in #98 , and upgrading to SAF would be lots of work, so I think that should go into its own separate issue.


nitehu:
Sorry I'm not sure why circleci failed on the tests... they are working for me :( Pls help

ogarcia:
Sorry for delay, too much work :-(

@nitehu you can view circleci output?
Anyway I attach it -> [circleci_output.txt](https://github.com/ultrasonic/ultrasonic/files/5116404/circleci_output.txt)

ogarcia:
It fails in same step:
```
      - run:
          name: unit-tests
          command: |
            ./gradlew ciTest testDebugUnitTest
            ./gradlew jacocoFullReport
            bash <(curl -s https://codecov.io/bash)
```
[circleci_output.txt](https://github.com/ultrasonic/ultrasonic/files/5117071/circleci_output.txt)

nitehu:
Thank you, I can see the logs but there isn't any code on the stack trace which would be inside Ultrasonic or something that it references...
I have no clue why it suddenly started to fail, we haven't changed anything related.
I'm trying to fix it by this [link to a similar error](https://github.com/gradle/gradle/issues/5184)

nitehu:
Ok, adding the missing jacoco exclude and updating the test libraries helped. Not sure what caused these errors but now it works.

PR #300, opened 2020-09-11
Remove all PullToRefresh dependencies

Maxmystere:
There were some odd behavior on views that had a "Album Card" on top of it coming from the PullToRefresh code so since androidx.swiperefreshlayout exist I replaced all dependencies to this package

Tests were made on modified views on a _Oneplus 8 with Android 11_ and a _Pixel 3a API 29 VM_

There are 2 TODO to check
- One for unknown legacy code
- One for a feature need for tchat view (replace scroll with button in Top App Bar ?)


ogarcia:
Hi @Maxmystere 

Great PR. Will you check the TODOs?

Maxmystere:
Hi @ogarcia !

For what I saw with the TODOs, the legacy code seems useless now and I couldn't test the tchat view since my subsonic server doesn't support tchat but the refresh behavior could be move to a button in the top-menu bar if necessary

ogarcia:
Ok. Do you want to add the button?

Maxmystere:
Okay I will try

ultrasonic/src/main/java/org/moire/ultrasonic/activity/ChatActivity.java
```diff
@@ -50,8 +53,6 @@ protected void onCreate(Bundle bundle)
 		setContentView(R.layout.chat);
 
 		refreshChatListView = findViewById(R.id.chat_entries_refresh);
-		// TODO Fix not able to pull from bottom
-		//refreshChatListView.setMode(Mode.PULL_FROM_END);
```
Maxmystere: Fixed by adding button in ChatView

ultrasonic/src/main/java/org/moire/ultrasonic/activity/SelectAlbumActivity.java
```diff
@@ -104,33 +104,6 @@ public void onRefresh() {
...
-					}
-				}
-			}
-
-			@Override
-			public void onScroll(AbsListView view, int firstVisibleItem, int visibleItemCount, int totalItemCount) {
-			}
-		});
-		*/
-
```
Maxmystere: Legacy Code when using PullToRefresh

Maxmystere (commented):
TODO to check

Maxmystere:
Need Translation

ogarcia:
@nitehu as I will be out until tuesday, if you want to test this PR, but I think that will work OK

nitehu (approved):
Review seems to be OK, I will execute some tests too...

nitehu:
@Maxmystere I've found a few errors during testing:

- In the Chat view, when I try to swipe to refresh, I get the following exception:

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.moire.ultrasonic.debug, PID: 16584
    java.lang.ClassCastException: androidx.appcompat.view.menu.ActionMenuItemView cannot be cast to android.view.MenuItem
        at org.moire.ultrasonic.activity.ChatActivity$4.onRefresh(ChatActivity.java:123)
        at androidx.swiperefreshlayout.widget.SwipeRefreshLayout$1.onAnimationEnd(SwipeRefreshLayout.java:195)

```

- A few activities are missing their bottom parts, e.g. the SelectAlbum Activity has a "more" arrow at the bottom and the chat activity has an Edit Text field to enter the chat message, these are missing after the PR. Could be because the SwipeRefresh has a height of "match_parent"

Also, could you please sign your commits? It is required as per the contribution guidelines. Unfortunately I can't think of a better solution to do this than setting up the signature and committing everything again, but I'm not a Git guru... @ogarcia have you got any better ideas?

Maxmystere:
I merged everything in one signed commit

- For simplicity i removed the swipe to refresh since the pull orientation is not good anymore and has been replaced with a button
-  I fixed all the layout_height and layout_width to match the old PullToRefresh layout

Commit history can still be seen [here](https://github.com/Maxmystere/ultrasonic/commits/maxdev)

nitehu:
Thank you very much!
Everything looks fine now.
The chat view refreshes itself in every few seconds anyway so I think too that the button will be enough.

ogarcia:
Merged! Will wait to #302 to upload a new release :wink:

Maxmystere:
Found a regression in the SelectArtistActivity, when clicking on the select_folder header the function [`folderButton.showContextMenu()`](https://github.com/ultrasonic/ultrasonic/blob/251e575421dd808083605d391deddf83987c27a0/ultrasonic/src/main/java/org/moire/ultrasonic/activity/SelectArtistActivity.java#L136) now return false and doesn't show the menu
I couldn't find a reason of it not working anymore

nitehu:
You won't believe this: in onCreateContextMenu the menuInfo.position has changed for this item from 1 to 0...
When I change `else if (info.position == 1)` to 0, it works. I can't imagine why :)
Created #305 for this

ogarcia:
The Artist List is a big mess. Maybe is a good moment to fix #304 :angel:

Maxmystere:
I guess it was because old PullToRefresh put another layout inside the list

Issue #84, opened 2017-11-23
Automate publishing the app


Tapchicoma:
Would be nice if new version of the app can be automatically published by CI.

We can use following plugin to do it: https://github.com/Triple-T/gradle-play-publisher
It supports Google Play Store and F-Droid (see #83).

Automatic publishing can be done on every merge to master branch (?)

ogarcia:
Is a good idea, but I think that is better use tags than every merge into master. Every new tag with version number is published in Google Play and F-Droid.

ogarcia:
This can be close for now. Only in play store must be published by hand, but prefer this method since Google continuosly change things in the store.

PR #302, opened 2020-09-19
Updated UI and Storage of Server Settings

### LINKED ISSUE #90: Server settings
Rethink how app shows and manages server settings.

Problems so far:
1. App shows modal dialogue for each field that user wants to edit.
2. App saves server settings in `SharedPreferences`, better to save them into DB.
3. List of servers are added to the app settings.

This issue is an epic issue. Particular changes should be done in a separate issues.

### LINKED ISSUE #156: Server settings mockups
This is only examples that we may use.

Personally i love the Yatse concept:

Server List
1: Servers list
2: Add new server button

The dropdown menu may be optional, but is a great idea for move server on list
Dropdown menu
The options 3 and 4 not apply in our case.

Add server window only ask for basic configuration:
Add Server

And sends you to *Edit Server* with two tabs
Edit Server Basic
Edit Server Advanced

In this case most of options of advanced view don't apply in Ultrasonic, but is a idea.

The basic settings can be equal:
- Name of connection
- Server Address
- Server port -> This setting may be in advanced under a option named _Use non standard port_?
- Usename
- Password
- Server enabled or disabled

The advanced settings can be:
- LDAP Support
- Default Jukebox
- Connection type (http or https) if user select https then show the _self-signed_ option.
- Future options...

As I say, this is only a possible view of server settings and it may not match the final version.

### LINKED ISSUE #217: If server name is not defined in server settings then it not appear in server selector
## Problem description

As has happened in the issue #216, if server name is not defined in server settings then it not appear in server selector.

### Steps to reproduce

1. Go config.
2. Add a server.
3. Set server URL, username and password, leave name in blank.
4. Test connection, it works fine.
5. Go back to main screen and try to select this server. Cannot because no name.

## Posible fixes

We can set server name as mandatory field or use server URL in case of server name is not defined. I think that second option is better, but maybe is more difficult to implement that the first one.

nitehu:
This PR is related to the Server Settings Milestone.

It contains a complete redesign of the Server Settings and Server Selection screens, and moves the storage of the settings from the Preferences to a Room database.
Also during the redesign the update of Koin to the latest version was necessary.

Fixes #217:
- The Name field has no extra function now, if it is empty when migrated to the database, it will be displayed as empty on the UI.
- However, after the migration the user can't create servers with empty names anymore. If the name is left empty, it will be filled with the Hostname part of the Url by default, as suggested in #217 

Fixes #156:
- The UI was redesigned to be something like Yatse's, as suggested. The servers can be edited and selected from a similar list, they have their own icon (for now signaling simply online/offline) and context menus for moving, editing and deleting. The list has a fancy FAB for new server creation :) .
- The New/Edit Server form is created with Material components, using separate categories for authentication and advanced settings. I didn't create separate "Setup Wizard" and "Advanced Settings Tab" like in Yatse, as Ultrasonic has so few settings that can be present on a single screen for better convenience.
- The Test Connection button was moved here as in Yatse.

Fixes #90:
- The storage of the Server Settings was moved to a Room SQLite databse. The migration is done the first time the application runs, it is hopefully unnoticable (the old settings aren't deleted yet from the Preferences if the user wants to downgrade version for some reason)

I think this PR will complete the **Server Settings Milestone**, as #84 shouldn't be there IMHO (please check)

Executed tests:
- Emulators: Galaxy Nexus API17, Pixel 2 API 27
- Phones: Samsung Nexus S with Android 4.4, Galaxy Note 9 with Andoid 10

Known problems:
- I did a "stress-test" too by pressing and filling everything like crazy, and one time I managed to stuck Ultrasonic in a state where the Jukebox was stuck On, and only restarting Ultrasonic fixed it. I'm not sure if this behaviour was present previously, but it is highly possible. I couldn't reproduce it.
- I've tried to test this extensively, however testing the migration part is very time-consuming. Any help is welcomed.

As this is a big PR, I will rebase it if Maxmystere's get accepted first (or I can help rebase that to this)


ogarcia:
WOW! Great work! I'm AFK until tuesday.

ogarcia:
@nitehu I do the tests and work fine, but I found a big problem, it lost old server settings. We must keep the old configuration if we do not want to be killed by negatives in the stores :joy:
I will test and merge Maxmystere's PR to that you can work in a _clean environment_.

nitehu:
Lost the old settings? That shouldn't happen :(
Did you have anything special in there? Can you try to reproduce it maybe, when you have some time?
It migrated the settings for me every time I tested, but it seems this needs much more testing.

ogarcia:
I can reproduce. I do the following:
1. Remove app
2. Install this version (with new UI), open it but do nothing.
3. Install current develop branch (without new UI) and configure a server.
4. Install this version and voilá, no settings.

I think that this is caused because you have a empty database and don't try to import the old settings. Maybe add a check like, if I have the new settings database but is empty, import old settings. What do you think?

nitehu:
You're right! The migration is only done at the first start (which is persisted in a preference). I didn't want to check for migration every time Ultrasonic starts (afraid of the performance impact), but it seems that it will be necessary.

I'll get back with a solution.

nitehu:
I've found the Jukebox problem, the good news is that it was in Ultrasonic before the PR. In the old version, when you delete a server from the settings, the other servers after it are moved back in Util.removeInstanceName. There the moved servers will get PREFERENCES_KEY_JUKEBOX_BY_DEFAULT = true as the default value if it is not set to false explicitly.
Hopefully migrating to the DB will solve this too.
I'll execute some more tests to be safe.

ogarcia:
Ok. When you consider that is mature to merge, tell me

nitehu:
I think everything is fixed now that I could find. I've updated the startup to check the servers in the DB with a "COUNT(*)", that should be reasonably fast even on SQLite. At least Ultrasonic doesn't seem to be any slower on my old Nexus S...

This can be merged now if you're OK with it.

ogarcia:
Merged!

@nitehu If a string changes you don't need to rename it. Rename string main.welcome_text that to main.welcome_text_new is unnecessary because we are using Transifex that marks changed strings as obsoleted/changed. :wink:

If you modify a string only need do the change in main file (and in translations that you can do, of course).

nitehu:
Oh wow, I didn't know Transifex does that! Thank you, next time I'll just simply change them.

nitehu:
@ogarcia please do not create a release yet, I've found another issue with the server indexes. :(
It only happens if there were some empty servers in the preferences for the user which doesn't get migrated, but I still want to fix it.

ogarcia:
No problem. I'm changing all icons to vectors and it takes a while :wink:

PR #318, opened 2020-09-29
Implemented Bluetooth connection settings

### LINKED ISSUE #266: Playback is paused when any bluetooth device disconnects in case multiple are connected
## Problem description

Hi, thanks for developing this app, I use it a lot.
Playback nicely pauses when disconnecting a bluetooth device, but apparently there is no check if the device doing the real playback is being disconnected. Playback is paused when any bluetooth device is disconnected.

### Steps to reproduce

The real use case is the combination of my mi band and bluetooth speaker. The mi band connects and reconnects a lot.
But for easier simulation just connect for example a bluetooth speaker and a laptop simultaneously to the android phone.
Start playback in Ultrasonic via the bluetooth speaker, while still connected to the laptop via bluetooth.
Then go to bluetooth settings in android and disconnect the laptop.

Result: Playback is paused.
Expected result: playback should not be paused, only if the real playback device should disconnect, in this case, the bluetooth speaker.
Result in case of mi band: it reconnects a lot, (so that's a problem too, not for this app of course), so playback is paused a lot of times. 

## System information

 * **Ultrasonic version**: *2.6.2*
 * **Android version**: *9, lineageos 16.0*
 * **Device info**: *Oneplus One*

## Additional notes

I'm willing to test any changes to the app.
If you need log files I can get them for you.
I also tested the behavior in other apps, f.e. VLC and they just continued playing.

nitehu:
Fixes #266 , and a few related issues too.

- There was a problem in Ultrasonic where it didn't detect A2DP connections / disconnections - it was subscribed for an old intent. This is fixed now, A2DP events are processed correctly.
- As mentioned in #266, there may be Bluetooth devices which are not detected as A2DP devices, so Ultrasonic paused itself when any Bluetooth devices disconnected. This caused problems for the users. 
  The solution is a new Setting with three values: 
  - pause playback on any device disconnection
  - pause playback only when an A2DP device is disconnected
  - do not pause playback on bluetooth events.
  The default "sane" value for this setting is pause on A2DP devices, which I think most users will use.
- Ultrasonic was able to resume playback on wired headset connection, however this didn't work for Bluetooth headsets. There is a new setting to fix this. Now Ultrasonic can be set up to resume on any bluetooth device connection, resume on A2DP devices, or do not resume at all. The default value is the latter, it won't resume without changing this setting.
- There were a few Intents registered to the MediaPlayerService in the AndroidManifest, but the handling of these were removed earlier. The solution is a new BroadcastReceiver which is registered to handle these Intents. Now at last Ultrasonic is able to resume and pause playback on Bluetooth events even from cold start.

### DIFF
--- a/ultrasonic/src/main/res/values-es/strings.xml
+++ b/ultrasonic/src/main/res/values-es/strings.xml
@@ -397,11 +397,11 @@
     <string name="settings.image_loader_concurrency_12">12</string>
     <string name="albumArt">Caratula del Álbum</string>
     <string name="common_multiple_years">Múltiples años</string>
-    <string name="settings.playback.resume_on_bluetooth_device">Resume when a Bluetooth device is connected</string>
-    <string name="settings.playback.pause_on_bluetooth_device">Pause when a Bluetooth device is disconnected</string>
-    <string name="settings.playback.bluetooth_all">All Bluetooth devices</string>
-    <string name="settings.playback.bluetooth_a2dp">Only A2dp devices</string>
-    <string name="settings.playback.bluetooth_disabled">Disabled</string>
+    <string name="settings.playback.resume_on_bluetooth_device">Reanudar al conectar un dispositivo Bluetooth</string>
+    <string name="settings.playback.pause_on_bluetooth_device">Pausar al desconectar un dispositivo Bluetooth</string>
+    <string name="settings.playback.bluetooth_all">Todos los dispositivos Bluetooth</string>
+    <string name="settings.playback.bluetooth_a2dp">Solo dispositivos de audio (A2DP)</string>
+    <string name="settings.playback.bluetooth_disabled">Deshabilitado</string>
 
     <string name="permissions.access_error">Ultrasonic no puede acceder a la caché de los ficheros de música. La ubicación de la caché se restableció a la ruta predeterminada.</string>
     <string name="permissions.message_box_title">Atención</string>
--- a/ultrasonic/src/main/res/values-es/strings.xml
+++ b/ultrasonic/src/main/res/values-es/strings.xml
@@ -397,6 +397,12 @@
     <string name="settings.image_loader_concurrency_12">12</string>
     <string name="albumArt">Caratula del Álbum</string>
     <string name="common_multiple_years">Múltiples años</string>
+    <string name="settings.playback.resume_on_bluetooth_device">Resume when a Bluetooth device is connected</string>
+    <string name="settings.playback.pause_on_bluetooth_device">Pause when a Bluetooth device is disconnected</string>
+    <string name="settings.playback.bluetooth_all">All Bluetooth devices</string>
+    <string name="settings.playback.bluetooth_a2dp">Only A2dp devices</string>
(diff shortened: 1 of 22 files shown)


ultrasonic/src/main/res/values/strings.xml
```diff
@@ -400,6 +400,11 @@
     <string name="albumArt">albumArt</string>
     <string name="common_multiple_years">Multiple Years</string>
     <string name="settings.server_address_unset" translatable="false">http://example.com</string>
+    <string name="settings.playback.resume_on_bluetooth_device">Resume when a Bluetooth device is connected</string>
+    <string name="settings.playback.pause_on_bluetooth_device">Pause when a Bluetooth device is disconnected</string>
+    <string name="settings.playback.bluetooth_all">All Bluetooth devices</string>
+    <string name="settings.playback.bluetooth_a2dp">Only A2dp devices</string>
```
ogarcia: I think that better string is:
*Only audio (A2DP) devices*
More users don't know what is A2DP

ultrasonic/src/main/res/values-es/strings.xml
```diff
@@ -397,6 +397,12 @@
     <string name="settings.image_loader_concurrency_12">12</string>
     <string name="albumArt">Caratula del Álbum</string>
     <string name="common_multiple_years">Múltiples años</string>
+    <string name="settings.playback.resume_on_bluetooth_device">Resume when a Bluetooth device is connected</string>
+    <string name="settings.playback.pause_on_bluetooth_device">Pause when a Bluetooth device is disconnected</string>
+    <string name="settings.playback.bluetooth_all">All Bluetooth devices</string>
+    <string name="settings.playback.bluetooth_a2dp">Only A2dp devices</string>
+    <string name="settings.playback.bluetooth_disabled">Disabled</string>
+
```
ogarcia, line 405: For do copy paste :joy:

```
<string name="settings.playback.resume_on_bluetooth_device">Reanudar al conectar un dispositivo Bluetooth</string>
<string name="settings.playback.pause_on_bluetooth_device">Pausar al desconectar un dispositivo Bluetooth</string>
<string name="settings.playback.bluetooth_all">Todos los dispositivos Bluetooth</string>
<string name="settings.playback.bluetooth_a2dp">Solo dispositivos de audio (A2DP)</string>
<string name="settings.playback.bluetooth_disabled">Deshabilitado</string>
```

ogarcia (changes requested):
This afternoon do tests with my devices ;-)

The only that I change is the A2DP text to be more clear

ultrasonic/src/main/res/values/strings.xml
nitehu: You're right, I was pondering on this too... fixed

ultrasonic/src/main/res/values-es/strings.xml
nitehu: copied 😊

ogarcia (approved):

ogarcia:
I think that can be merged without problem :1st_place_medal:

Issue #317, opened 2020-09-29
[Feature Request] Dark Theme (Black Amoled)


awsms:
Hi, I wanted to know if you were planning to add a completely black theme at some point?
Thanks for all the great work on this app btw!

awsms:
Any chance you could also add a _true black_ mode, like [this](https://i.imgur.com/Ehhzjmr.png)?

awsms:
Any chance you could also add a _true black_ mode, like [this](https://i.imgur.com/Ehhzjmr.png)?

Not that I dislike this current black/cyan theme, which looks really neat! But I was specifically a looking for a black amoled mode :-)

ogarcia:
@awsms the black theme is based in `Theme.MaterialComponents`. Is the one that [Google recommends to OLED Screens][1] (see second paragraph).

[1]: https://material.io/develop/android/theming/dark

PR #326, opened 2020-10-04
Add Material Black Theme

Maxmystere:
Feature for #317

Tested on Oneplus 8 with android 11

### DIFF
--- a/ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
+++ b/ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
@@ -153,7 +153,25 @@ public class Util
 	public static String getTheme(Context context)
 	{
 		SharedPreferences preferences = getPreferences(context);
-		return preferences.getString(Constants.PREFERENCES_KEY_THEME, "dark");
+		return preferences.getString(Constants.PREFERENCES_KEY_THEME, Constants.PREFERENCES_KEY_THEME_DARK);
+	}
+
+	public static void applyTheme(Context context)
+	{
+		String theme = Util.getTheme(context);
+
+		if (Constants.PREFERENCES_KEY_THEME_DARK.equalsIgnoreCase(theme) || "fullscreen".equalsIgnoreCase(theme))
+		{
+			context.setTheme(R.style.UltrasonicTheme);
+		}
+		else if (Constants.PREFERENCES_KEY_THEME_BLACK.equalsIgnoreCase(theme))
+		{
+			context.setTheme(R.style.UltrasonicTheme_Black);
+		}
+		else if (Constants.PREFERENCES_KEY_THEME_LIGHT.equalsIgnoreCase(theme) || "fullscreenlight".equalsIgnoreCase(theme))
+		{
+			context.setTheme(R.style.UltrasonicTheme_Light);
+		}
 	}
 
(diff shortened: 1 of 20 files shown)


ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
```diff
@@ -153,7 +153,25 @@ public static boolean isLockScreenEnabled(Context context)
...
+		}
+		else if (Constants.PREFERENCES_KEY_THEME_BLACK.equalsIgnoreCase(theme))
+		{
+			context.setTheme(R.style.UltrasonicTheme_Black);
+		}
+		else if (Constants.PREFERENCES_KEY_THEME_LIGHT.equalsIgnoreCase(theme) || "fullscreenlight".equalsIgnoreCase(theme))
+		{
+			context.setTheme(R.style.UltrasonicTheme_Light);
+		}
 	}
```
Maxmystere, line 175: Move all function to one Util function

ultrasonic/src/main/res/values/themes.xml
```diff
@@ -1,5 +1,63 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+    <style name="UltrasonicTheme.Black" parent="Theme.MaterialComponents">
```
Maxmystere, line 3: Using material design component (Amoled)

Maxmystere:
I don't know how translation should be handled so i put default english for strings that I can't translate

nitehu:
That's exactly how translations should be handled ;)
Thanks for your work! We'll check this and get back to you.

ogarcia (changes requested):
This is not more necessary. We remove the translations check because it has not sense and is better to transifex translators. Please remove [this commit](https://github.com/ultrasonic/ultrasonic/pull/326/commits/cd4b1239fca83deb15ab55aa7c2441d2fc1560b4) (or simply remove untranslated strings).

And sorry for the inconvenience :pray:

nitehu (approved):

nitehu:
I think this will be my default theme 😊

ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
```diff
@@ -153,7 +153,25 @@ public static boolean isLockScreenEnabled(Context context)
...
 		SharedPreferences preferences = getPreferences(context);
-		return preferences.getString(Constants.PREFERENCES_KEY_THEME, "dark");
+		return preferences.getString(Constants.PREFERENCES_KEY_THEME, Constants.PREFERENCES_KEY_THEME_DARK);
+	}
+
+	public static void applyTheme(Context context)
+	{
+		String theme = Util.getTheme(context);
+
+		if (Constants.PREFERENCES_KEY_THEME_DARK.equalsIgnoreCase(theme) || "fullscreen".equalsIgnoreCase(theme))
```
Maxmystere, line 163: Btw I do not know what this "fullscreen" check is inside that function

ogarcia: Seems legacy code. There is a lot of legacy code in this app... :shrug:

ogarcia (approved):

PR #329, opened 2020-10-07
Fixed Opus file handling and offline file name bug

### LINKED ISSUE #227: Offline mode does not show Opus files.
## Problem description

UltraSonic can play Opus files, but they do not appear when in offline mode.

### Steps to reproduce

1. Play/cache a `.opus` file from a server.
1. Switch to offline mode
1. Track/file does not appear anywhere.

## System information

 * **Ultrasonic version**: 2.6.0
 * **Android version**: 8.1
 * **Device info**: Nexus 5X

### LINKED ISSUE #328: File loaded to cache directory from computer is unplayable
## Problem description

When I upload a file from my computer to the Ultrasonic cache directory, Ultrasonic is unable to play the file in offline mode.
Maybe related to #223 

### Steps to reproduce

1. Load a file into Ultrasonic cache directory
2. Take Ultrasonic to offline mode and select the file to play
3. Ultrasonic will stay in "Buffering" state indefinitely, and won't play the file. In the logcat, it continuously tries to connect to 127.0.0.1 but throws an exception.

## Additional notes

The problem is in the FileUtil.getSongFile function. 
- This function will give a new name to the downloaded song
- The song won't be present with the new name
- The DownloadFile will try to download it, which will fail in Offline mode.

Actually, I'm not even sure why does the getSongFile tries to find out the name of the file, when a perfectly usable file is present in the "Path" property already.

nitehu:
Fixes #227 
- This is a straightforward fix, "opus" extension was missing from the list of audio file extensions which Ultrasonic recognizes.

Fixes #328
- Checks if the file in the "Path" property already exists, and won't rename it if it does.

### DIFF
--- a/ultrasonic/src/main/java/org/moire/ultrasonic/util/FileUtil.java
+++ b/ultrasonic/src/main/java/org/moire/ultrasonic/util/FileUtil.java
@@ -51,7 +51,7 @@ public class FileUtil
 	private static final String TAG = FileUtil.class.getSimpleName();
 	private static final String[] FILE_SYSTEM_UNSAFE = {"/", "\\", "..", ":", "\"", "?", "*", "<", ">", "|"};
 	private static final String[] FILE_SYSTEM_UNSAFE_DIR = {"\\", "..", ":", "\"", "?", "*", "<", ">", "|"};
-	private static final List<String> MUSIC_FILE_EXTENSIONS = Arrays.asList("mp3", "ogg", "aac", "flac", "m4a", "wav", "wma");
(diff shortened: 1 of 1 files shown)


nitehu:
I wasn't too happy with the first solution to the file name problem. This happens very rarely (I think, Ultrasonic is not really an offline player, and it isn't a good idea to leave its cache folder as your general music directory anyway, as Ultrasonic will delete files if it thinks so), and checking every file's existence when a big playlist is added may reduce performance on weaker phones.

PR #333, opened 2020-10-13
Fixed Subsonic API version detection by storing minimum version in Server Settings

### LINKED ISSUE #253: Ultrasonic send p auth instead of t+s auth when using API 1.13.0
I've added a workaround for this in Ampache but here's the issue.

By default it seems Ultrasonic supports 1.7.0 auth. I've recently added 1.13.0 support to ampache which uses token authentication.

When connecting to Ampache with ultrasonic i will sometimes get a 1.7.0 request instead of the token request which will fail to connect.

It seems to be mostly around ping and getlicense
```
"GET /rest/ping.view?u=%user%&c=Ultrasonic&f=json&v=1.7.0&p=enc:%AUTH% HTTP/1.1" 200 4064 "-" "okhttp/3.10.0"
"GET /rest/getLicense.view?u=%user%&c=Ultrasonic&f=json&v=1.13.0&p=enc:%AUTH% HTTP/1.1" 200 595 "-" "okhttp/3.10.0
```
Once it has had an error connecting it will work again but Ultrasonic should probably pick an auth method and stick with it.

I think setting auth for p or t+s based on "Enable support for LDAP users" setting would probably be an easy way to fix this.

nitehu:
Fixes #253 

- Added a new server property to settings DB: minimumApiVersion
  This is filled with the API version returned in the first Ping result when the server is first contacted.
  Afterwards this version is retrieved from the server settings to be used as the initial version to contact the server.
  The servers will only see Ultrasonic's minimum API version (1.7) at the very first Ping (with possibly wrong authentication in case of e.g. Ampache). After that in subsequent requests Ultrasonic will use the authentication according to the stored server version. 
  (This also means that users degrading their server version may need to re-enter the server details in Ultrasonic)
  
- Fixed MainActivity's server setting change detection
  It didn't react if one property of a server changed, only noticed the selected ID change. This is now fixed.
  
- Fixed MD5 Salt to be a different value per request as the Subsonic API specifies
 
- Fixed the description of the "LDAP" server setting
  I always forgot what it actually does until I looked it up in the code, I hope the new text is more clear.


ogarcia:
> * (This also means that users degrading their server version may need to re-enter the server details in Ultrasonic)

I think is better that when user enter in server settings for edit them, simply delete this entry to force check again. So the user only need enter in settings and press back, save or test connection.

What do you think?

nitehu:
Good idea, I'll add it.

I thought about this server version degrading and it seems it shouldn't cause much problems for the users, as only the initial request will use the new version, Ultrasonic will detect the correct version afterwards. So the users will probably notice this only if someone degrades from 1.13 to pre-1.13, where the authentication changes.

nitehu:
- Modified the EditServerActivity to set the minimumApiVersion to null when the user enters to edit a server.

- I wanted to detect the version before any call to the server, as the first Ping is not guaranteed to succeed (e.g. because of a network error).
  Also calling the Ping from the MainActivity was a temporary solution. Now before any call the version will be determined first.
  This was way easier (and nicer) to do in Kotlin, so the RESTMusicService was refactored from Java.
  
- This broke the OfflineMusicService, which for some reason was a subclass of the RESTMusicService. 
  This subclassing caused some problems before, so now they're entirely separated, and the Offline service won't call REST requests unintentionally on 127.0.0.1 anymore.
  
This together is now a bigger change than I expected, so it's probably a good idea to test this during the weekend.

nitehu:
Tested what I could and everything seems to work.

ogarcia:
This weekend will do intensive review :smiling_face_with_three_hearts:

PR #345, opened 2020-11-04
Fixed too small icons and margins

### LINKED ISSUE #342: Notification control icons should be bigger
## Problem description

The control icons in the notifications are too small, sometimes it is hard to touch the one the user wants. They should be bigger, or there should be more padding / margin between them. It is especially noticeable on the small notification.

## Additional notes

As I know this (or a similar problem) was fixed before, but it seems the new vector icons are smaller than the previous ones, and that causes this problem again.

### LINKED ISSUE #344: Increase media_buttons size
## Problem description
The media buttons are very small on my device.
I would like the media buttons (epescially media_play / pause) to be displayed bigger, similar to the way f.e. spotifiy is displaying it.

Imho, these are the most used control and the are are mesely 5% of screen height, making them hard to access.

The play button in ultrasonic is ~96px height, while f.e. in spotify, it is 172px high (values for my device) and more prominently spaced.

Screenshot (04 11 2020 14_21_23)

Screenshot (04 11 2020 14_31_38)

## Additional notes

If nescecary, the cover-display size could be traded for that. It is very big anyway.

## System information

 * **Ultrasonic version**: *2.15.0*
 * **Android version**: *11*, RP1A.201005.004
 * **Device info**: *Google, Pixel 2*

BTW. Thank you all very much for this app, its a real joy to see a great app with active developers!

nitehu:
Fixes #344 , fixes #342 

- Changed icon sizes from the default 24dp to 32dp in widgets, album_buttons, small and large notifications and mini now playing.
- Added margins in now playing and song list items (the drag icon was too close to the scroll bar)
- Made the media buttons larger, the play/pause/stop is the largest. Now it is similar to how spotify looks. It looks a bit funny now, I'm not sure if I should just get used to it, or now it's too large. What do you think, @ogarcia ?

Tested it on some screen resolutions, generally it looks good.

### DIFF
--- a/ultrasonic/src/main/res/layout/album_buttons.xml
+++ b/ultrasonic/src/main/res/layout/album_buttons.xml
@@ -9,7 +9,8 @@
     <ImageView
         android:id="@+id/select_album_select"
         android:layout_width="0dp"
-        android:layout_height="wrap_content"
+        android:layout_height="32dp"
+        android:scaleType="fitCenter"
         android:layout_weight="1"
         android:src="?attr/select_all"
         android:visibility="gone" />
@@ -17,7 +18,8 @@
     <ImageView
         android:id="@+id/select_album_play_now"
         android:layout_width="0dp"
-        android:layout_height="wrap_content"
+        android:layout_height="32dp"
+        android:scaleType="fitCenter"
         android:layout_weight="1"
         android:src="?attr/media_play"
         android:visibility="gone" />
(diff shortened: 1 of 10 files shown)


ogarcia:
Let me test :wink: ...

ogarcia:
Merged! I do small change to make play icon even BIGGER!

nitehu:
👍 nice :D

PR #346, opened 2020-11-04
Improve player layout

ogarcia:
I think that is more clear now and much more beautiful in landscape.

@nitehu What do you think?


nitehu (approved):

nitehu:
Wow, it looks good! Like a video player :)

PR #356, opened 2020-11-29
Set target sdk to 29

ogarcia:
Check the new SDK with CI


ogarcia:
@nitehu I close and delete this change because need more work. I'll publish in beta channel the new release with this change does by hand, If you can look into it I would appreciate.

PR #358, opened 2020-11-29
Updated API to 29, fixed failing tests

### LINKED ISSUE #357: Fix tests to set target SDK to 29 (or upper)
For upload new releases to Google Play Store must need to [set the target SDK to 29](https://developer.android.com/distribute/best-practices/develop/target-sdk). I do it by hand to version 2.17.0, but we need fix the tests to set target to level 29 or upper.

nitehu:
Fixes #357 
It seems something was updated in the compiler and now it notices that getClassLoader() may return null. 
I think a not-null-assertion can safely be added, especially as the change only seems to affect the unit tests.


ogarcia:
I did several tests with API29 and no fail found. If you haven't found any problems either, you can merge this and I make a new version revision 2.17.1.

Issue #219, opened 2018-07-30
Fallback to file list when no track number tag found


4oo4:
## Problem description
I have a number of files with the tracks numbered vinyl-style (A1, A1, B2...), and Ultrasonic does not order them correctly in an album view. I don't think Airsonic parses them correctly, even though they are in the FLAC tags, when I do edit tags on the file the Track tag is blank. However, in Airsonic the files are still listed in the correct order, since the files start with the track number.

Is it possible to have Ultrasonic fallback to sorting files alphabetically when it can't find a track number? I think that's the behavior that the Airsonic web client does in that situation.

### Steps to reproduce
 1. Import files into Airsonic, with Track Number FLAC/ID3 tags as A1, A2...
 2. Browse to album in Airsonic, all tracks have -1 for the track number and are in random order

## System information

 * **Ultrasonic version**: 2.6.0
 * **Android version**: 8.1.0
 * **Device info**: OnePlus 5T

Cheers

Tapchicoma:
Would be very helpful if you can provide several files with such tags. Can send them via gitter.

4oo4:
Can do, I'll try and do that tonight.

**EDIT:** Linked to an example in Gitter, this is what it looks like in the app:

![screenshot_ultrasonic_20180801-201309](https://user-images.githubusercontent.com/15698835/43556881-8e180754-95f1-11e8-89a7-7a2fcddc7e81.png)

4oo4:
I was poking around to see how hard this would be to implement, and I'm thinking that this might be the culprit:

https://github.com/ultrasonic/ultrasonic/blob/master/ultrasonic/src/main/java/org/moire/ultrasonic/util/EntryByDiscAndTrackComparator.java

Then one could potentially look here to look at the file listing?

https://github.com/ultrasonic/ultrasonic/blob/master/ultrasonic/src/main/java/org/moire/ultrasonic/util/FileUtil.java

Would that be a good way of approaching it?

I haven't really worked with Java/Kotlin much, but would be willing to take a stab at it if my hunch is correct.

tzugen:
Is this still happening?

4oo4:
Sorry, forgot this was still open. Since I've been stuck at home I haven't used Ultrasonic for a while, but I will test this once I get my Airsonic server recovered from a hardware failure and let you know.

Cheers

ogarcia:
[...]

PR #364, opened 2020-12-18
Fall back to path when comparing tracks

lbonn:
Tracks will be sorted by ascending path if neither album, disc and track id can help.

This is helpful for browsing a loose collection of tracks tagged under the same album.

Similar to #219

I've tested the changes in an emulator, with navidrome as a server.

### DIFF
--- a/ultrasonic/src/main/java/org/moire/ultrasonic/util/EntryByDiscAndTrackComparator.java
+++ b/ultrasonic/src/main/java/org/moire/ultrasonic/util/EntryByDiscAndTrackComparator.java
@@ -18,6 +18,8 @@ public class EntryByDiscAndTrackComparator implements Comparator<MusicDirectory.
 		Integer trackY = y.getTrack();
 		String albumX = x.getAlbum();
 		String albumY = y.getAlbum();
+		String pathX = x.getPath();
+		String pathY = y.getPath();
 
 		int albumComparison = compare(albumX, albumY);
 
@@ -33,7 +35,14 @@ public class EntryByDiscAndTrackComparator implements Comparator<MusicDirectory.
 			return discComparison;
 		}
 
-		return compare(trackX == null ? 0 : trackX, trackY == null ? 0 : trackY);
+		int trackComparison = compare(trackX == null ? 0 : trackX, trackY == null ? 0 : trackY);
+
+		if (trackComparison != 0)
+		{
+			return trackComparison;
+		}
+
+		return compare(pathX == null ? "" : pathX, pathY == null ? "" : pathY);
 	}
 
 	private static int compare(long a, long b)


ogarcia:
@lbonn thanks for your PR but you must sign your commit, see [CONTRIBUTING](CONTRIBUTING.md).

lbonn:
@ogarcia I tried to follow the insctructions and used `--signoff` to acknowledge the contribution will be licensed as GPLv3.

Do you mean cryptographically signed?
Or the "Signed-off-by:" must be followed by a real name?

Maxmystere:
image
You should have this when commit is signed
(Github auto sign commits when done with the web text editor if you don't wanna bother yourself with recommiting from repo)

ogarcia:
@lbonn [This](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/about-commit-signature-verification)

I just update CONTRIBUTING to clarify this point :wink:

lbonn:
Thanks for the clarification, I will try to do that by the end of the week-end.

But then I realized I might have taken the wrong approach, and that the sorting feature seems to be intended as a last resort for broken ordering from the server and that it should be first fixed there if possible.
For example, it also sorts "Most Played" and "Recently Played" albums by their names (though, we could probably fix that as well)

But we can still merge this if you think it's a good addition.

nitehu (approved):
I think the change is OK, if it fixes the ordering for some users... With my test servers the track comparision usually has a result, so it won't go comparing the paths unless I change the result in debug.

lbonn:
Ok, re-pushed after a gpg signature.

ultrasonic/src/main/java/org/moire/ultrasonic/util/EntryByDiscAndTrackComparator.java
```diff
@@ -33,7 +35,14 @@ public int compare(MusicDirectory.Entry x, MusicDirectory.Entry y)
...
 
-		return compare(trackX == null ? 0 : trackX, trackY == null ? 0 : trackY);
+		int trackComparison = compare(trackX == null ? 0 : trackX, trackY == null ? 0 : trackY);
+
+		if (trackComparison != 0)
+		{
+			return trackComparison;
+		}
+
+		return compare(pathX == null ? "" : pathX, pathY == null ? "" : pathY);
```
ogarcia, line 45: I have a doubt with this. `compare` is a function that compares two _longs_, but `pathX` and `pathY` are strings. I think that this is a `TypeError`

lbonn: It's overloaded, there is a `private static int compare(String a, String b)` on line 53

PR #365, opened 2020-12-21
Do not sort special collections

lbonn:
It does not make sense to sort:

- random songs
- recently added albums
- recently played albums
- most played albums
- top rated albums
- random albums

As the order returned by the server is more relevant in these cases


nitehu (approved):
Nice find, I'd say this was almost a bug. I didn't even notice it just by using Ultrasonic, the returned lists seemed to be fairly random :)

ogarcia:
Great PR. I'll AFK until next week. But it will be merged :wink:

PR #366, opened 2020-12-24
Send "Accept" headers in Subsonic API calls

lbonn:
Some picky servers only responds when this is present (curl sends it by default for example).

I noticed this when trying to a navidrome server running in development mode.

### DIFF
--- a/core/subsonic-api/src/main/kotlin/org/moire/ultrasonic/api/subsonic/SubsonicAPIClient.kt
+++ b/core/subsonic-api/src/main/kotlin/org/moire/ultrasonic/api/subsonic/SubsonicAPIClient.kt
@@ -70,7 +70,11 @@ class SubsonicAPIClient(
                 .addQueryParameter("c", config.clientID)
                 .addQueryParameter("f", "json")
                 .build()
-            chain.proceed(originalRequest.newBuilder().url(newUrl).build())
+            var newRequest = originalRequest.newBuilder()
+                .url(newUrl)
+                .header("Accept", "*/*")
+                .build()
+            chain.proceed(newRequest)
         }
         .addInterceptor(versionInterceptor)
         .addInterceptor(proxyPasswordInterceptor)


deluan:
Even though I think it is a good thing to add the `Accept` header in the requests, I don't think that is causing any issues (at least not with Navidrome). I just tested it and it answers correctly when the `Accept` header is missing:

```shell
$ curl -H 'Accept:' 'https://demo.navidrome.org/rest/getAlbumList.view?f=json&u=demo&p=demo&v=1.12&c=postman&type=newest' 
```
Response: 
```json
{"subsonic-response":{"status":"ok","version":"1.16.1","type":"navidrome","serverVersion":"0.39.0 (f2a8308)","albumList":{"album":[{"id":"c502ccf2d9928dda656cce3f2fbf239e","parent":"1e9fa068b94cd5ed844bd1b73ecb6acc","isDir":true,"title":"Fairytale","name":"Fairytale","album":"Fairytale","artist":"Natasha Beller","year":2018,"genre":"Trip-Hop","coverArt":"bd086ed63aa854bf5ab9e54e822cb04e","duration":1838,"playcount":20,"created":"2020-12-03T14:20:29.451808083Z","artistId":"1e9fa068b94cd5ed844bd1b73ecb6acc","userRating":4,"songCount":9,"isVideo":false},{"id":"a17d8b958d64d6dafbcf157ff458a747","parent":"03b645ef2100dfc42fa9785ea3102295","isDir":true,"title":"netBloc Vol. 42: Live, The Universe \u0026 Everything","name":"netBloc Vol. 42: Live, The Universe \u0026 Everything","album":"netBloc Vol. 42: Live, The Universe \u0026 Everything","artist":"Various Artists","year":2013,"coverArt":"c3688eff1a2454fbe1d464a8b79b0d49","duration":2951,"playcount":3,"created":"2020-12-03T14:08:42.564237076Z","artistId":"03b645ef2100dfc42fa9785ea3102295","songCount":11,"isVideo":false}]}}}
```

lbonn:
I did not have problem with regular Navidrome but only when running it with `make dev`: https://www.navidrome.org/docs/developers/dev-environment/. I think it's because the requests to the backend are proxied by a nodejs server launched by react's development mode.

All api calls then return 404 if they don't include this header.

deluan:
Ah, the issue is different: `make dev` run a [nodejs proxy in front of the server](https://create-react-app.dev/docs/proxying-api-requests-in-development/) and only forward requests that accepts JSON. From the docs:
> The development server will only attempt to send requests without text/html in its Accept header to the proxy.

To avoid this you have two options:
1) When starting with `make dev`, configure your Subsonic client with port 4633 (instead of 4533). This is the direct port of the server when running in `dev` mode, bypassing the nodejs proxy
2) If you are only testing the Subsonic API, start Navidrome with `make server`, and it does not start the UI js server, so requests to port 4533 are directly handled by Navidrome's server

Having said that, I'm all for adding `Accept` header to the request.

PS: Sorry for hijacking this PR with information for other project, but I think it is relevante to understand the context of the proposed PR

ogarcia:
Sorry, but add an accept */* header has no sense, it's the same that not send it.

lbonn:
@ogarcia that's a good point actually, it is quite redundant like this.

And if we add `Accept: application/json`?

nitehu:
I think `Accept: application/json` can't be added in a similar way as above (part of all requests), as the servers may send pictures or streams as a response for some requests.

Issue #391, opened 2021-03-13
Add "Select Folder" to album view


SaintDubious:
## Problem description

My server has my music nicely organized into different catalogs. When I go into the Artist View there is a drop down at the top called "Select Folder" that lets me filter by which catalog I want to see. The Album View doesn't give me this option, instead forcing me to see albums from all of my catalogs.

I'd like to add the same "Select Folder" option to the Album View

### Steps to reproduce

When not in "Offline Mode" (ie make sure you have a subsonic server set up), from the "Ultrasonic Main" menu select "Albums." Notice that there is no "Select Folder" option.

## System information

 * **Ultrasonic version**: 2.18.0
 * **Android version**: 11
 * **Device info**: Pixel 3

## Additional notes

I'm a software developer and I'd like to contribute to this fix. I've been reading the code and I see that Artist View has been converted to Kotlin while the other views have not. I was thinking a necessary first step would be to convert the Album View to Kotlin. I was gonna start working on that now and try to create a PR for the work. After that's done I can see about adding the "Select Folder" option.

SaintDubious:
The above PR solves this issue, though there's still more work to clean it up and search for bugs. Hopefully I'll get it nailed down soon.

SaintDubious:
I got it nailed down :) this one is ready for review.

PR #392, opened 2021-03-13
Convert select album fragment

SaintDubious:
I'd like to be able to add a "Select Folder" option to the top of the Album List (see issue: https://github.com/ultrasonic/ultrasonic/issues/391). Since the contributing guidelines say all new code has to be in Kotlin, I decided to start by converting the `SelectAlbumFragment` to Kotlin.

I used the automatic tools to do the initial conversion, then I worked through the issues highlighted by the Android Studio IDE. From there I tested by just clicking on everything I could possible find in the UI.

This is a fairly large change that's hard to diff. I'm happy to followup with tests etc as suggested.


SaintDubious:
Further testing notes: I added a breakpoint to every function and then ran in debug mode, clearing out breakpoints as I hit them. This at least gives some confidence that every function has been run (obviously not 100% coverage, but a start). I couldn't get the following to run. If someone can provide instructions on how to test them I'm happy to try it out:

- getPodcastEpisodes
- getShare

SaintDubious:
The gradle detektCheck failed with a "Forbidden Comment" on a comment with TODO. I have no idea what that's about, the comment was there before I started. I changed the spelling to "To do" and it seems happy.

nitehu:
Thank you very much for your work! 

Ktlint and detekt can be annoyingly pedantic at times. They only check kotlin files, so "TODO:" was accepted when it was in java. Please use "TODO" without the semicolon in kotlin. Generally I think it is a bad idea to change it to anything else like "to do" as the IDEs (or somebody running a manual search) won't find it that way. Android Studio for example will accept and list "TODO" and "TODO:" too. 

I also think it is a good practice to keep "zero warning policy" in the source files as much as possible, especially as Android Studio usually has good suggestions which will help to keep a readable code. For example, there are a few "Name shadowed" warnings remained in the code, where a local variable has the same name as a function parameter. I'm sure this wasn't intentional, maybe just the kotlin converter did it, but you should check these as the code is harder to read and understand this way. Please also check if the other warnings make sense and easy to fix.

As I see you've made a thorough testing, I'll check if I can help with something. On one of my servers I may have a few shares so I'll try to check getShare too.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
```diff
@@ -0,0 +1,1180 @@
...
+    private var deleteButton: ImageView? = null
+    private var moreButton: ImageView? = null
+    private var playAllButtonVisible = false
+    private var shareButtonVisible = false
+    private var playAllButton: MenuItem? = null
+    private var shareButton: MenuItem? = null
+    private var showHeader = true
+    private val random: Random = SecureRandom()
+
+    private val mediaPlayerControllerLazy = KoinJavaComponent.inject(
```
nitehu: This is not a Lazy Init anymore, so can be renamed to mediaPlayerController
Also please use the kotlin way for DI with Koin here ("by inject()"), please see SelectArtistFragment for reference

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
```diff
@@ -0,0 +1,1180 @@
...
+            )
+            selectAll(false, false)
+        }
+    }
+
+    private fun getMusicDirectory(refresh: Boolean, id: String?, name: String?, parentId: String?) {
+        setTitle(this, name)
+
+        object : LoadTask() {
+            @Throws(Exception::class)
```
nitehu: I think "@Throws" has no meaning in a kotlin private code. This annotation is for functions that may get called from Java. Please check if they can be removed everywhere else in this file.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
```diff
@@ -0,0 +1,1180 @@
...
+
+            if (sortableCollection() && Util.getShouldSortByDisc(context)) {
+                Collections.sort(entries, EntryByDiscAndTrackComparator())
+            }
+
+            var allVideos = true
+            var songCount = 0
+
+            for (
+                (_, _, isDirectory, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, isVideo)
```
nitehu: Wow maybe it's just me but this seems to be ugly, just use `for (entry in entries)`

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
```diff
@@ -0,0 +1,1180 @@
...
+            val duration = Util.formatTotalDuration(albumHeader.totalDuration)
+
+            val durationView = header!!.findViewById<TextView>(R.id.select_album_duration)
+            durationView.text = duration
+
+            return header
+        }
+    }
+
+    companion object {
```
nitehu: Maybe these can be simple private members of the class, what do you think? Companion objects usually are necessary when they contain something public.

nitehu (commented):
Could you please check these comments?

SaintDubious:
I got it to compile without warnings and I wanted to push that as a distinct commit for easy review. 

The "shadow" warnings were all from when an argument had a given name but then within the function a lambda (are they called "lambdas" in Kotlin?) was created for a callback function, and the lambda had an argument of the same name. This same shadowing existed in the Java code (see [example here](https://github.com/ultrasonic/ultrasonic/blob/develop/ultrasonic/src/main/java/org/moire/ultrasonic/fragment/SelectAlbumFragment.java#L130)) but maybe in Java the variable isn't captured? Or maybe Java just complains differently? In any event, I did some clunky renaming by prepending "the" in a few places. Since the compiler won't catch it if I miss a replacement please triple check this part.

The other warnings of note were two places where the old Java code was checking for `null` when Kotlin guarantees it can't be `null`. I'm an old C++ guy, so not checking for NULL terrifies me, but I did verify that the function that's being called is Kotlin and not Java, so I guess it's safe?

I also set it to "TODO" ;)

Lastly, I did the usual test by setting breakpoints and hitting them all, these are the ones I couldn't get to:
- the 2 `null` checks. Yikes
- the click listener for getSongsForGenre

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
SaintDubious: Ha! Score one for the auto-conversion tools :)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
SaintDubious: Please make sure I did this correctly. I tried to turn the others into `inject()` but it didn't work for Java reasons I don't understand.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
SaintDubious: All of the `@throws` I could find were in reference to the `LoadTask` class, which is private. So I believe it isn't possible for anything else to call them. I removed them all.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
SaintDubious: yeah, that "companion" thing is new to me. I think that's how the Java->Kotlin conversion tools handle statics. I'm not sure why these things were static in the first place. I made them private and everything works just fine.

SaintDubious:
That's all from me tonight. Thanks @nitehu for the great feedback. I'm glad to know that code reviews are taken seriously, it boosts my confidence in this project.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
nitehu: I think the function was static because it was a pure function (didn't use instance properties, just its parameters) and those are usually marked as static in java and similar languages. Anyway, kotlin doesn't have static and the companion wasn't necessary (some use the companion in kotlin to store pure functions in it, but I'm not sure if that's a good practice - it looks ugly especially when you try to access them from java).

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
nitehu: Yep you did everything fine! You can change the others too, just remove the `.value` part, e.g.:
```
    private val videoPlayer: VideoPlayer by inject()
    ...
    videoPlayer.playVideo(entry)

```
This is necessary because the DI works differently in java and in kotlin, in java it uses Lazy Initialization (with the Lazy<T> class), which receives its value when the `.value` is first accessed. In kotlin, the injected objects don't need to be Lazy<T>, it uses some kotlin magic instead in `by inject()` to return a type T directly...

SaintDubious: Worked great! I'm not sure what went wrong the last time I tried, probably a simple typo. Anyway, it works now

SaintDubious:
Just a small change tonight, I switched over to using `inject()`. One concern on testing is that I've lost my ability to pin. Not just on this branch, but even on master. I tried clearing all my data, but no luck. Please test this on your side and if it fails for you let me know and I'll spend some more time trying to sort it out.

nitehu:
Sorry I had to merge in a quick fix, I hope it won't cause any problems for you. It wasn't related to this PR, but will help in the testing of the Podcasts functionality.

SaintDubious:
No problem at all. I merged it in and did another round of testing. I figured out how to get pinning working as well, so that's now re-tested. I'm happy with the state of this PR, let me know if you spot anything in your own tests.

nitehu (approved):
Review OK

nitehu:
I've tested the shares and podcasts, no problem there. All other functions seem to work as expected too.

@SaintDubious unfortunately I've just noticed there will be one more thing you need to solve: as you can see below, merging is blocked because you need to sign your commits before they can be accepted. This is mentioned in the contribution guidelines, and you can find a link there too to a page which describes how to do that. 
I think existing commits can't be signed, but you can force push a signed commit to this branch, which can be then merged.

SaintDubious:
I hope I did that right! For the first one I used the key, but forgot the `--signoff` on the license. The second has both.

SaintDubious:
Okay so this gpg thing clearly threw me for a loop. This PR still says "Merging is blocked" I assume because there are commits in it that aren't signed? Thankfully the diff is trivial, so I created a second PR, but when I merged `develop` that went in as unsigned and that PR was blocked. So I made a third one that I hope will do the trick. Hopefully you can merge one of these and kill the other.

Sorry for the hassle.

PR #405, opened 2021-03-24
Refactor LocalMediaPlayer to Kotlin, fix #390

### LINKED ISSUE #390: Crashes when playing audio
## Problem description

Sometimes I expirience random crashes.

### Steps to reproduce

Not yet clear

## System information

 * **Ultrasonic version**:  Latest on F-Droid
 * **Android version**: Lineage 17.1
 * **Device info**: *not provided

## Additional notes

Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0 in tid 29016 (Thread-8), pid 28948 (oire.ultrasonic)
03-10 13:22:14.794 29019 29019 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
03-10 13:22:14.794 29019 29019 F DEBUG   : LineageOS Version: '17.1-xxxxxxxxxxxxxxxxxxxxxxxx'
03-10 13:22:14.794 29019 29019 F DEBUG   : Build fingerprint: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
03-10 13:22:14.794 29019 29019 F DEBUG   : Revision: '0'
03-10 13:22:14.794 29019 29019 F DEBUG   : ABI: 'arm64'
03-10 13:22:14.796 29019 29019 F DEBUG   : Timestamp: 2021-03-10 13:22:14+0100
03-10 13:22:14.796 29019 29019 F DEBUG   : pid: 28948, tid: 29016, name: Thread-8  >>> org.moire.ultrasonic <<<
03-10 13:22:14.796 29019 29019 F DEBUG   : uid: 10149
03-10 13:22:14.796 29019 29019 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
03-10 13:22:14.796 29019 29019 F DEBUG   : Cause: null pointer dereference
03-10 13:22:14.796 29019 29019 F DEBUG   :     x0  0000007666a65900  x1  00000076634dfb90  x2  0000000000000001  x3  0000000012f04d94
03-10 13:22:14.796 29019 29019 F DEBUG   :     x4  0000000000000000  x5  0000007672a60a00  x6  0000007672a60a00  x7  0000000000000000
03-10 13:22:14.796 29019 29019 F DEBUG   :     x8  0000000000000006  x9  0000007673035804  x10 0000007673038228  x11 0000000000000041
03-10 13:22:14.796 29019 29019 F DEBUG   :     x12 00000076cd05b0a8  x13 00000076cd05b0b0  x14 00000076d8b562cc  x15 0000000000000000
03-10 13:22:14.796 29019 29019 F DEBUG   :     x16 000000767303d5e8  x17 0000007759e1a630  x18 0000007661eec000  x19 0000000000000000
03-10 13:22:14.796 29019 29019 F DEBUG   :     x20 00000076634dfb90  x21 0000007666a65900  x22 0000000000000002  x23 00000000fffffff9
03-10 13:22:14.796 29019 29019 F DEBUG   :     x24 0000000012f04d84  x25 00000076729147ec  x26 00000000ffffffe0  x27 0000000012f04d94
03-10 13:22:14.796 29019 29019 F DEBUG   :     x28 00000076634e1020  x29 00000076634dfb70
03-10 13:22:14.797 29019 29019 F DEBUG   :     sp  00000076634dfb50  lr  000000767303823c  pc  0000007759e1a64c
03-10 13:22:15.260 29019 29019 F DEBUG   : 
03-10 13:22:15.260 29019 29019 F DEBUG   : backtrace:
03-10 13:22:15.260 29019 29019 F DEBUG   :       #00 pc 000000000000f64c  /system/lib64/libutils.so (android::RefBase::decStrong(void const*) const+28) (BuildId: 6212421b5da38101f8412487a7e3e28e)
03-10 13:22:15.260 29019 29019 F DEBUG   :       #01 pc 0000000000004238  /system/lib64/libaudioeffect_jni.so (android_media_AudioEffect_native_getParameter(_JNIEnv*, _jobject*, int, _jbyteArray*, int, _jbyteArray*)+616) (BuildId: 58219892f2590f2e29f34622141ee178)
03-10 13:22:15.260 29019 29019 F DEBUG   :       #02 pc 000000000029efd4  /system/framework/arm64/boot-framework.oat (art_jni_trampoline+196) (BuildId: 600a26b6da00607997f3350c09c9e3dec3da8b47)
03-10 13:22:15.260 29019 29019 F DEBUG   :       #03 pc 0000000000136334  /apex/com.android.runtime/lib64/libart.so (art_quick_invoke_stub+548) (BuildId: a0de32d5cd8b6ef7386adb27fa34113b)
03-10 13:22:15.261 29019 29019 F DEBUG   :       #04 pc 0000000000145064  /apex/com.android.runtime/lib64/libart.so (art::ArtMethod::Invoke(art::Thread*, unsigned int*, unsigned int, art::JValue*, char const*)+244) (BuildId: a0de32d5cd8b6ef7386adb27fa34113b)
03-10 13:22:15.261 29019 29019 F DEBUG   :       #05 pc 00000000002e343c  /apex/com.android.runtime/lib64/libart.so (art::interpreter::ArtInterpreterToCompiledCodeBridge(art::Thread*, art::ArtMethod*, art::ShadowFrame*, unsigned short, art::JValue*)+384) (BuildId: a0de32d5cd8b6ef7386adb27fa34113b)
03-10 13:22:15.261 29019 29019 F DEBUG   :       #06 pc 00000000002de334  /apex/com.android.runtime/lib64/libart.so (bool art::interpreter::DoCall<false, false>(art::ArtMethod*, art::Thread*, art::ShadowFrame&, art::Instruction const*, unsigned short, art::JValue*)+928) (BuildId: a0de32d5cd8b6ef7386adb27fa34113b)
03-10 13:22:15.261 29019 29019 F DEBUG   :       #07 pc 00000000005a3b48  /apex/com.android.runtime/lib64/libart.so (MterpInvokeDirect+400) (BuildId: a0de32d5cd8b6ef7386adb27fa34113b)

tzugen:
While refactoring I found three weird lines, which could be actually bugs, or just hacks without comment. 
I annotated them with #FIXME , please check them out...


tzugen:
What is the idea with checkstyle and too long lines if the method signature is that long?
I could abbreviate the parameters, but that doesn't make things more readable or understandable, no?

`    private inner class BufferTask(private val downloadFile: DownloadFile, private val position: Int) : CancellableTask() {`

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/LocalMediaPlayer.kt
```diff
@@ -0,0 +1,832 @@
...
+
+    @Synchronized
+    fun playNext() {
+        if (nextMediaPlayer == null || currentPlaying == null) return
+
+        val oldPlayer = mediaPlayer
+        mediaPlayer = nextMediaPlayer!!
+
+        // FIXME: Why is this being done?
+        nextMediaPlayer = oldPlayer
```
nitehu: I think the original logic was that there are two instances of initialized MediaPlayers, while one is playing the other is being prepared. Here the two are swapped (using a temp variable).
This seems to be no longer necessary as the nextPlayer is always recreated as a new MediaPlayer... 
You may remove it and see if it breaks anything.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/LocalMediaPlayer.kt
```diff
@@ -0,0 +1,832 @@
...
+
+        setCurrentPlaying(nextPlaying)
+        setPlayerState(PlayerState.STARTED)
+
+        // FIXME: Why is currentPlaying passed here and not nextPlaying?!
+        attachHandlersToPlayer(mediaPlayer, currentPlaying!!, false)
```
nitehu: I think after the setCurrentPlaying() call the currentPlaying and nextPlaying will be the same.
Change this as you like (or just add a comment for anyone reads the code next)

tzugen:
The static analysis fails only because of the //FIXME comments I added ;)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/LocalMediaPlayer.kt
```diff
@@ -0,0 +1,831 @@
...
+        if (!Util.isLockScreenEnabled(context)) {
+            clearRemoteControl()
+            return
+        }
+
+        if (remoteControlClient == null) {
+            remoteControlClient = createRemoteControlClient()
+        } else {
+            // FIXME: This looks like a hack. Why is it needed?
+            audioManager.unregisterRemoteControlClient(remoteControlClient)
```
nitehu, line 311: This seems to be a workaround, and I'm not sure for what reason it was added. I've found this nugget of wisdom by googleing a bit: 
"You must register your RemoteControlDisplay every time when the View which displays metadata is shown to the user. This is because 4.2.2 and lower versions support only one RemoteControlDisplay, and if system will decide to register it's own RCD, your RCD will be unregistered automatically."
This may explain it, if it is true for RemoteControlClients too.
As Ultrasonic supports older Androids, I think it is enough to add a comment here for anyone reads the code next.

nitehu (changes requested):
Wow nice work, thank you!
I've added a few comments for the FIXMEs, please check if those make sense to you.
Otherwise, the review was OK.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/LocalMediaPlayer.kt
tzugen: Seems to work ok, if removed :)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/LocalMediaPlayer.kt
tzugen: Yep, i changed it to be more understandable.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/LocalMediaPlayer.kt
tzugen: ok!

tzugen:
Ok, thanks for the comments.

I uncovered another bug, probably in MediaPlayerService #406 
Afterwords I've been listening to music most of the day and it is stable so far.

nitehu:
Ok so I think I've found an issue. When Ultrasonic is started, it starts the playback immediately, which is not how it woked before.

It is regarding the change in MediaPlayerControllerImpl (doPlay -> play). Play calls bufferAndPlay, which starts a BufferTask, and when that's completed, the playback is started instantly (practically this happens when Ultrasonic is started)
Originally doPlay had a "start" parameter which was used to signal whether autostart is necessary, at Ultrasonic startup this is false by default, so the playback wasn't started.

Could you please check this?

tzugen:
> Ok so I think I've found an issue. When Ultrasonic is started, it starts the playback immediately, which is not how it woked before.
> 
> It is regarding the change in MediaPlayerControllerImpl (doPlay -> play). Play calls bufferAndPlay, which starts a BufferTask, and when that's completed, the playback is started instantly (practically this happens when Ultrasonic is started)
> Originally doPlay had a "start" parameter which was used to signal whether autostart is necessary, at Ultrasonic startup this is false by default, so the playback wasn't started.
> 
> Could you please check this?

Do you know the exact steps to reproduce? I can't seem to do it. The restore method in MediaPlayerControllerImpl gets passed an AutoPlay boolean, and this boolean is passed down the chain...

tzugen:
Ah, i got it reproduce.

tzugen:
The interesting thing is that in BufferTask doPlay() was called also before with autoPlay set to true, not respecting the value from restore(). 
Funny that that popped up now :shrug:

nitehu:
Thank you, it seems to be OK now. I'll test it for 1-2 days to see if everything else is fine too.

nitehu (approved):

nitehu:
I think there's another issue, suddenly Ultrasonic was throwing an exception when I pressed play, and playback wouldn't start.
The exception was
```
08:19:11.396: I LocalMediaPlayer Preparing media player 
08:19:11.398: W LocalMediaPlayer Media player error
java.io.IOException: setDataSourceFD failed.: status=0x80000000
	at android.media.MediaPlayer._setDataSource(Native Method)
	at android.media.MediaPlayer.setDataSource(MediaPlayer.java:1356)
	at android.media.MediaPlayer.setDataSource(MediaPlayer.java:1339)
	at android.media.MediaPlayer.setDataSource(MediaPlayer.java:1278)
	at android.media.MediaPlayer.setDataSource(MediaPlayer.java:1239)
	at android.media.MediaPlayer.setDataSource(MediaPlayer.java:1204)
	at org.moire.ultrasonic.service.LocalMediaPlayer.doPlay(LocalMediaPlayer.kt:521)
	at org.moire.ultrasonic.service.LocalMediaPlayer.access$doPlay(LocalMediaPlayer.kt:42)
	at org.moire.ultrasonic.service.LocalMediaPlayer$BufferTask.execute(LocalMediaPlayer.kt:728)
	at org.moire.ultrasonic.util.CancellableTask$1.run(CancellableTask.java:84)
 java.io.IOException: setDataSourceFD failed.: status=0x80000000

```
Of course as soon as I've attached the debugger it fixed itself, so I'm trying to reproduce it now. I was listening to some music in the car when it froze, so maybe it was related to start/stop by bluetooth disconnect/reconnects.

nitehu:
I think I've found what causes #406 .
LocalMediaPlayer is a singleton in the DI, so the same instance is used everywhere in Ultrasonic while it runs.
That's why in the original file the mediaPlayer was initialized in the onCreate function.
When you moved the initialization in the constructor, that broke, since there is a single instance of LocalMediaPlayer and the constructor will only run once.
Please fix this (e.g. by moving back the initialization of mediaPlayer to the onCreate)

tzugen:
Ok, I understand the Singleton concept now. 
But I am not sure if moving the init of mediaPlayer to onCreate is the right solution.

It feels semantically wrong that the class has an onDestroy function, but is never actually destroyed, so the method onCreate and onDestroy are actually more an init() and cleanup() method.

I haven't wrapped my head around it completely, but there are two approaches that look saner to me:
* Rename the onCreate/onDestroy methods to something semantically more meaningful, and init/release mediaPlayer there.
* Maybe better?: Make  LocalMediaPlayer a scoped instance, so that it follows the lifecycle of the service ( https://insert-koin.io/docs/reference/koin-core/scopes/ )

nitehu:
I just suggested to move it back to the onCreate because that is how it was in the Java version, and sometimes there are legacy things which will fall apart quickly if we try to change them (this doesn't mean that they absolutely shouldn't be changed, but they will require a lot more work and testing, and these kind of things better done in small, focused steps).

For example, if you try to change LocalMediaPlayer to a scoped instance, you'll find that it is referenced from MediaPlayerControllerImpl and Downloader, which are also singletons, and the UI uses them so can't be scoped, and this PR will just get bigger and bigger. 

I however absolutely agree that the onCreate/onDestroy naming is a bit misleading as they suggest Android manages its lifecycle somehow, so it would be nice to change them to something more understandable.

And of course many kudos to you if later when you have time (and patience) you'll try to cut order into this legacy spaghetti by refactoring the media player parts in one or more separate PRs 😊

tzugen:
Ok cool, thanks :)
I will rename the methods, but make them work like before.

I have opened a new bug (#409 ) because I think these issues would be worth discussing more,
but not as part of this PR

tzugen:
Ok, renamed. I assign a new MediaPlayer instance after release, instead of making the property nullable, because in a future PR, I would have probably reverted it again. Hope that is ok as well :)

ogarcia (approved):
LGTM. @nitehu if you don't see more inconvenience you can merge this pull request

nitehu:
Thank you! Just that one exception above bothers me. 
We are going on a trip on the weekend and we'll use Ultrasonic a lot, if I still can't reproduce it I'll merge this PR.

PR #410, opened 2021-04-02
Introduce the SelectMusicFolderView

### LINKED ISSUE #391: Add "Select Folder" to album view

SaintDubious:
See: https://github.com/ultrasonic/ultrasonic/issues/391

This PR brings the ability to select and filter Albums by a Music Folder on the server. Previously it was only possible to do this with the Artist List View.

I refactored the old ArtistRowAdapter::HeaderViewHolder into a self contained SelectMusicFolderView. The idea is that this little View shows the currently select folder, and can manage a popup that shows all the folders on the server and lets you pick one. When one is picked it will perform a callback to let the calling class handle the update. I then put this back into the SelectArtistFragment AND introduced it into the SelectAlbumFragment. 


ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
```diff
@@ -117,6 +126,24 @@ class SelectAlbumFragment : Fragment() {
...
+                false, requireContext()
+            ),
+            activeServerProvider.getActiveServer().musicFolderId,
+            { _, selectedFolderId ->
+                if (!ActiveServerProvider.isOffline(context)) {
+                    val currentSetting = activeServerProvider.getActiveServer()
+                    currentSetting.musicFolderId = selectedFolderId
+                    serverSettingsModel.updateItem(currentSetting)
+                }
+                artistListModel.refresh(refreshAlbumListView!!)
```
SaintDubious: One note on this line. I really don't like that the SelectAlbumFragment now knows about the artist list. However the problem is that the CachedMusicServer offers no way to clear out the cache without changing the URL. So when a new folder is selected, the artists from the old folder are still in the cache. This is the only way I could find to force them to update. 

Suggestions on a better solution would be greatly appreciated.

SaintDubious:
I figured out how to restrict the new header to only the album list view and not the permutation like show recent, show starred, show most played, etc. It may make its way to those views as well, but I'm less familiar with them and not really sure it makes sense. Like if you just want to see favorites, it probably makes sense to not filter your favorites? Anyway, this is now ready for submission

ultrasonic/src/main/kotlin/org/moire/ultrasonic/view/SelectMusicFolderView.kt
```diff
@@ -0,0 +1,81 @@
...
+) : RecyclerView.ViewHolder(
+    LayoutInflater.from(context).inflate(
+        R.layout.select_folder_header, root, false
+    )
+) {
+    private val folderName: TextView = itemView.findViewById(R.id.select_folder_2)
+    private val layout: LinearLayout = itemView.findViewById(R.id.select_folder_header)
+
+    init {
+        if (selectedFolderId != null) {
```
nitehu, line 42: IMHO this needs an else case where the folderName is set to "All folders". Currently the text will be empty when the fragment is first started.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
```diff
@@ -735,6 +763,11 @@ class SelectAlbumFragment : Fragment() {
     private fun getAlbumList(albumListType: String, albumListTitle: Int, size: Int, offset: Int) {
 
         showHeader = false
+        showSelectFolderHeader = !isOffline(context) &&
```
nitehu: If you check the SelectArtistFragment, there should be another condition to show the header: `!Util.getShouldUseId3Tags`
I think this is because if this setting is turned on, then another subsonic API will be used, which doesn't handle folders.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
nitehu: I think this can be solved in the checkSettingsChanged() in CachedMusicService. The musicFolderId can be stored similarly as the Url, and the cache can be deleted automatically if that changes.

ultrasonic/src/main/res/layout/select_folder_header.xml
```diff
@@ -24,7 +24,7 @@
         a:orientation="vertical" >
 
         <TextView
-            a:id="@+id/select_artist_folder_1"
+            a:id="@+id/select_folder_1"
```
nitehu: I know this was bad before, but I think if you change the name anyways, it can be renamed to something meaningful e.g. select_folder_title and select_folder_name below or something similar.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/view/SelectMusicFolderView.kt
```diff
@@ -0,0 +1,81 @@
...
+                    break
+                }
+            }
+        }
+        layout.setOnClickListener { onFolderClick() }
+    }
+
+    private fun onFolderClick() {
+        val popup = PopupMenu(context, layout)
+        val MENU_GROUP_MUSIC_FOLDER = 10
```
nitehu: These private constants can be lifted before the "class" (after the imports) in kotlin, `private const val MENU_GROUP_MUSIC_FOLDER = 10`, and Anroid Studio will like it better :)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/view/SelectMusicFolderView.kt
```diff
@@ -0,0 +1,81 @@
...
+
+        var menuItem = popup.menu.add(
+            MENU_GROUP_MUSIC_FOLDER, -1, 0, R.string.select_artist_all_folders
+        )
+        if (selectedFolderId == null || selectedFolderId!!.isEmpty()) {
+            menuItem.isChecked = true
+        }
+        for (i in musicFolders.indices) {
```
nitehu: I know this was copied (probably the kotlin converter created it this way in the past), but it would be really more readable if it was changed to `musicFolders.forEachIndexed`

nitehu (changes requested):

nitehu:
> I figured out how to restrict the new header to only the album list view and not the permutation like show recent, show starred, show most played, etc. It may make its way to those views as well, but I'm less familiar with them and not really sure it makes sense. Like if you just want to see favorites, it probably makes sense to not filter your favorites? Anyway, this is now ready for submission

I'm not sure either, the Subsonic API must be checked to see what requests support the folders. Anyways I think this is OK for now as is, and if somebody finds a good use case for the folders elsewhere, it can be added relatively easily.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
SaintDubious: Thank You! This is an elegant solution, it works great.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
SaintDubious: It turns out that the other API, `getAlbumList2(...)`, actually does support it, but the argument was hard coded to `null`. I passed through the music folder id and it worked.

SaintDubious:
Thanks as always @nitehu for the review. Hopefully I fixed everything.

nitehu (approved):
Review OK. Lets test it for a few days to see if everything is fine.

nitehu:
I've found a strange issue... 
The problem seems to be that the SelectAlbumFragment never calls the getMusicFolders() with refresh = true. This is made worse by that the RestMusicService caches the music folder list to a file (I can't imagine why), so Ultrasonic won't even refresh the folder list when it starts. The list can only be refreshed it seems from the Artist fragment, because that calls with refresh = true by the SwipeRefresh.

I think a solution would be that if the selectFolderHeader would be initialized inside the updateDisplay function, where the refresh is available as a parameter. This way the SwipeRefresh would reload the folder list too.
@SaintDubious what do you think? Could you check this please?

SaintDubious:
Fantastic catch @nitehu . I didn't think to test adding new catalogs on my server, my bad. Sadly the fix is not so simple **and** my PR actually breaks the SelectArtistFragment as well. You can't refresh the music folder list in the `updateDisplay` function because that would be calling the network on the main UI thread, which causes an exception. The old code uses a `LiveData<>` object to hold the music folders, and a coroutine in `ArtistListModel` to load the data. It looks like on startup it just uses an empty list and makes the view say "All Folders" and then when the coroutine loads up the proper music folder list it quickly updates everything.

I think I'm starting to see how the pieces fit together, but I'll need to spend more time on this one. I'll update the title to mark this as **DO NOT MERGE** for now.

SaintDubious:
Okay, I think I have a handle on this one. A couple of changes here. The initial one is that the `SelectMusicFolderView` now loads up with "All Folders" as the only option, but supports a `setData(...)` function to provide it the selected folder and the full list of folders. Then I hooked into the pre-existing method of downloading music folders through the `SelectArtistFragment` to provide this data (not a very large change). I also copied that method into the `SelectAlbumFragment` such that it can now download the folders in a separate thread when `updateDisplay` is called.

While testing this I noticed that the `RESTMusicService::getIndexes` function was broken. It has a cache on the file system, but that cache doesn't take into account the passed in `musicFolderId`. So if the cache is created when with folder id '1' but then you switch to folder id '2' it will still return the contents of '1'. I imagine this was never a problem before because the only way to change the folder id was through the `SelectArtistFragment` which then had enough context to know to refresh everything. Anyway, to fix it I just made separate caches for each folder ID. 

Lastly, I actually changed some folder names on my music server and verified that the folder list updated when I refreshed, so it's working as expected.

I'm gonna remove the DO NOT MERGE in the name. These changes were big enough that another review is warranted.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/view/SelectMusicFolderView.kt
```diff
@@ -0,0 +1,89 @@
...
+import android.widget.PopupMenu
+import android.widget.TextView
+import androidx.recyclerview.widget.RecyclerView
+import org.moire.ultrasonic.R
+import org.moire.ultrasonic.domain.MusicFolder
+
+/**
+ * This little view shows the currently selected Folder (or catalog) on the music server.
+ * When clicked it will drop down a list of all available Folders and allow you to
+ * select one. The intended usage us to supply a filter to lists of artists, albums, etc
```
tzugen: us -> is

tzugen:
When I go to the "Recently Added albums", now only those albums from the collection that was last set in the Albums view are shown. Before all recent albums were shown. I think the previous behaviour was better.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
```diff
@@ -117,6 +137,29 @@ class SelectAlbumFragment : Fragment() {
             false
         )
 
+        selectFolderHeader = SelectMusicFolderView(
+            requireContext(), albumListView!!,
+            { selectedFolderId ->
+                if (!ActiveServerProvider.isOffline(context)) {
+                    val currentSetting = activeServerProvider.getActiveServer()
+                    currentSetting.musicFolderId = selectedFolderId
+                    serverSettingsModel.updateItem(currentSetting)
```
tzugen, line 146: <s>Maybe don't save it as a global setting, as it will affect a bunch of other queries in views, were there is (and maybe should not) be a folder selector?</s>

SaintDubious:
Thank You @tzugen ! You were correct, the "Recently Added" was picking up the music folder filter even though it wasn't showing the header. I checked in a pretty unobtrusive fix for this one. I also re-verified all the other views to make sure they only filter by the folder if they allow you to select the folder. I also fixed my typo :)

tzugen (approved):
Yes, that looks like a nice place to put the fix :)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
```diff
@@ -735,6 +803,11 @@ class SelectAlbumFragment : Fragment() {
     private fun getAlbumList(albumListType: String, albumListTitle: Int, size: Int, offset: Int) {
 
         showHeader = false
+        showSelectFolderHeader = !isOffline(context) &&
```
tzugen: If the useID3 setting is basically a setting which determines, whether to show your music as it is stored hierachically on the file system, or based on its ID3 tags, then for consistency with the Artist View, where the folder selector is also not shown when ID3 is enabled, it should take the same conditions here, IMHO

ogarcia:
> When I go to the "Recently Added albums", now only those albums from the collection that was last set in the Albums view are shown. Before all recent albums were shown. I think the previous behaviour was better.

IMHO, the same behabiour must be in:
- Recently Played
- Most Played
- Top Rated
- Random
- Starred

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
SaintDubious: I'm of two minds about what's "proper behavior" but I think the consistency argument is solid. Currently the Artist view doesn't show the header, so the Album view shouldn't either. Thankfully it's trivial to change back and forth so I changed it.

SaintDubious:
> > When I go to the "Recently Added albums", now only those albums from the collection that was last set in the Albums view are shown. Before all recent albums were shown. I think the previous behaviour was better.
> 
> IMHO, the same behabiour must be in:
> 
> * Recently Played
> * Most Played
> * Top Rated
> * Random
> * Starred

Not very promising so far I'm afraid. The "Recently Played," "Most Played," and "Top Rated" all go through the same path, and I can get them to all pass the music folder ID to the server, but the server seems to ignore it. I'll investigate this one further, maybe there's a bug in ampache (the server I'm using).

"Random" has some promise, I'll work on this one now. I'd also like to have this one supported for my own use.

"Starred" view doesn't work for me. I filed an issue here https://github.com/ultrasonic/ultrasonic/issues/417 but I'm afraid until I get that fixed I won't be able to test if the drop down works :(

SaintDubious:
Sorry @ogarcia , no luck. I traced down the call stack far enough to see that the client was making the call to `getAlbumList` as specified [here](http://www.subsonic.org/pages/api.jsp#getAlbumList), but my test server ignored the music folder id for the ones you requested (even "Random").

How strongly do you feel about those other views? Would you be comfortable merging this as is and deferring that work? Or would you prefer I find some alternate test server and/or try to work on an ampache fix?

tzugen:
> Sorry @ogarcia , no luck. I traced down the call stack far enough to see that the client was making the call to `getAlbumList` as specified [here](http://www.subsonic.org/pages/api.jsp#getAlbumList), but my test server ignored the music folder id for the ones you requested (even "Random").
> 
> How strongly do you feel about those other views? Would you be comfortable merging this as is and deferring that work? Or would you prefer I find some alternate test server and/or try to work on an ampache fix?

Hi, I am pretty sure there is no problem. I just checked out the latest of your branch, and it all works correctly for me now.
When I go to Random or Starred, the list contains Entry from all Directories, even if I before selected a folder.

It makes also sense: For example in RESTMusicServiceKt, L497:
```
        val response = responseChecker.callWithResponseCheck { api ->
            api.getRandomSongs(
                size,
                null,
                null,
                null,
                null
            ).execute()
        }
```

The last null is the hardcoded musicfolderid.

I can see nothing that should stop a merge anymore 😁

nitehu:
This is really a minor thing, but triggers my OCD:
Screenshot 2021-04-12 173425
The "No media found" text should be at the bottom, what do you think? :)

nitehu (approved):

nitehu:
> Hi, I am pretty sure there is no problem. I just checked out the latest of your branch, and it all works correctly for me now.
When I go to Random or Starred, the list contains Entry from all Directories, even if I before selected a folder.

Yep I agree, I've also checked all pages and they work the same. Recently Added, Recently Played, Most Played, Top Rated, Random, Starred now all return results from all folders. IMHO it is meaningless to filter these by folder, and from the tests @SaintDubious made it seems the servers wouldn't support filtering either.

ogarcia:
> The "No media found" text should be at the bottom, what do you think? :)

Agree

ogarcia (approved):
LGTM, but need fix the "Not media found" to appear under folder selector

SaintDubious:
It's always the simple requests that get ya. Any help on this one would be greatly appreciated. That "No Media Found" TextView is listed in [the xml file](https://github.com/ultrasonic/ultrasonic/blob/develop/ultrasonic/src/main/res/layout/select_album.xml#L12) on top of the ListView that contains the new header, so that's why it's showing up on top. Seems reasonable to me to just put the TextView underneath the SwipeRefreshLayout, and that works, but then the text is pinned to the bottom of the screen. Now's where it gets frustrating. What I'd want to do is make the ListView just NOT fill up the whole screen when all it contains is the header. That would then let the TextView slide right up underneath it. However any attempt I make to set it to wrap_content or give it a zero height or anything else causes an uncaught exception at runtime. 

The only thing I can make work is if I give the TextView a weight of around 8, but that's just the hackiest possible solution, I'm sure it would look poor on some screens.

I'm giving up for the night :(

tzugen:
I would be happy to help,
but I won't have time until the weekend.

Let me know...

Also I thought that the whole SelectAlbumView needs a serious refactor,
in the direction that you already worked, and wanted to ask you if you are interested in working on this together?

@nitehu  @ogarcia  Maybe it would be also ok for the maintainers to merge this, if we make sure to take care of it in the next step? It would be great to merge this, so that I can rebase the other PRs etc..

nitehu:
The problem is that we can't use a simple text before/after the listView anymore, because the SwipeRefresh and the listView fills the screen. And rightly so, because we must be able to refresh the list even when there are no items in it.

The solution IMHO is that the "No media found" text should go inside the list, and the easiest way to solve that is to add a footer or add it inside the header.

I think it shouldn't take long to fix this, but if there are other unforeseen problems then maybe just remove the text altogether, I'm not sure we must write inside an empty list that the list is empty (and can add it back later after the refactor if necessary).

ogarcia:
> The solution IMHO is that the "No media found" text should go inside the list, and the easiest way to solve that is to add a footer or add it inside the header.

Agree.

tzugen:
More specifically, I plan to convert those ListViews to RecyclerViews anyways, and in this step it would make sense to make the display of "Nothing found" a thing that the RecyclerView takes care of if its empty.
 A fix now could possibly be just double work.

ogarcia:
If you all agree, we can merge this PR to continue work and wait to fix before publish the new release. Sounds good to you?

nitehu:
OK for me!

SaintDubious:
Thanks everyone! I will continue working on putting that label inside the list... that actually sounds like a way simpler solution, wish I thought of it last night.

@tzugen refactoring this one makes sense, but that's a big topic. I'd actually go one step further into exploring the UI a bit. The top level menu has "Ultrasonic Main", which is a way to filter by just about everything (including artist) and "Media Library", which is filter by artist. Why two options when one includes the other? Except when you're in offline mode when for some reason your only option is to filter by artist. 

UI aside, the SelectArtist and SelectAlbum fragments sort of do the same thing (show a list of music containers and let you pick) and yet the implementations are vastly different. It feels like there was once just SelectAlbum and then someone decided they were gonna refactor the artists into a new thing and then go back for more... but never got around to it.

In any event, I guess what I'm saying is that I agree with you, but I'm not up to such a huge task yet. Once I finish with this "no media found" label thing there are 2 other issues that stand out to me as more critical/immediate. But maybe once I get through my own wish list I will be more comfortable working on that big one with you :)

PR #419, opened 2021-04-14
Clean empty text view

SaintDubious:
A continuation and cleanup from https://github.com/ultrasonic/ultrasonic/pull/410

The "No media found" text was appearing stubbornly on top of the `SelectMusicFolderView`. This fix leaves the original TextView hidden (because some other screens still use it) and instead simply adds a second header to reproduce the message. 

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SelectAlbumFragment.kt
@@ -1118,12 +1118,11 @@ class SelectAlbumFragment : Fragment() {
             playAllButtonVisible = !(isAlbumList || entries.isEmpty()) && !allVideos
             shareButtonVisible = !isOffline(context) && songCount > 0
 
+            albumListView!!.removeHeaderView(emptyView!!)
             if (entries.isEmpty()) {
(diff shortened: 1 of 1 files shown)


SaintDubious:
@nitehu you're a super hero for suggesting this. Last night I spent like 2 hours beating my head into my keyboard in frustration trying to move the xml around. This solution was a 15 minute fix that just worked. Thank you!

nitehu (approved):

nitehu:
As @tzugen said, fixing this is unnecessary when the whole list will be refactored to recyclerWiew, however I can understand if it bugs you and want to complete your work.
Anyways, add a check please before adding the header to see if it was added before, the current solution behaves **very** strange when you refresh the list with a swipe 😊

SaintDubious:
As you say, I don't really like leaving my work unfinished. I guess feel free to merge or not depending on how the refactor goes.

This fix is a bit hacky, but counting how many headers there are is a bit brittle as there could be other kinds of headers that may or may not be there. I was hoping for some kind of function like `doesThisHeaderExist(...)` but I couldn't find one, so I just always remove it and maybe add it.

nitehu:
Yep, unfortunately the API lacks in handling this, e.g. the select folder header is also just added based on the count... We can live with this.

Issue #418, opened 2021-04-13
Regression?


tzugen:
## Problem description

Yesterday I had problems with unexpectedly stopping playback again. I couldn't look to deep into it yet, but it could have been a regression in LocalMediaPlayer or a an Adroid issue with MediaPlayer.  I won't have time to dig into it until the weekend, so it would be great if you didn't release a new version until then.

tzugen:
Do you want to skip 2.19 or close it?
If you want to release 2.19, you should cherry-pick this commit, as it fixes this bug:
https://github.com/ultrasonic/ultrasonic/pull/421/commits/427c81acc16c99d7de06f057398d27b4cb908c5c

ogarcia:
I think that better perform a new release 2.10.0 with all changes

PR #421, opened 2021-04-17
Modernize Service Notification

tzugen:
Use the native Android way of generating playback notifications, by using a "Media Style" and creating a Media Session.

Look how nice :)

Screenshot_20210417-195743 cleaned


tzugen:
I found the fix for #418 and justed added it here for reasons of laziness. But I can make a seperate PR if you want.

nitehu:
Wow nice work! This was also on my todo list, however I had run into some problems while planning to implement it... These were because I couldn't find out how more buttons can be added to the notification. (I think the max number is 4?)

- I miss the stars 😢(I can live without them but they were really convenient)
- Something must be done to stop the playback and dismiss the notification. This was done with the close ("X") button before, but now that is missing. Please take a look at Poweramp, this is one solution:
Screenshot 2021-04-18 101215

- On the emulator when the notification is dismissed, the music is just continues playback in the background, I couldn't reproduce it on real phones but it needs testing... I hope it is just an emulator thing, as Android specifies that when a service is running it must have a notification.
- Also, now this appeared on my Ultrasonic icon:
Screenshot 2021-04-18 101612
Poweramp doesn't do this so it can be removed somehow...

ogarcia:
Impressive work!

> * I miss the stars cry(I can live without them but they were really convenient)

About this, YouTube music has thumb up and thumb down buttons:
ytm

I think that add a one star button to mark as favorite can be a good idea. But, the five stars selection must burn in hell.

> * Also, now this appeared on my Ultrasonic icon:
>   Screenshot 2021-04-18 101612
>   Poweramp doesn't do this so it can be removed somehow...

Yes, and Ultrasonic has not this before

tzugen:
Hi :)

Regarding the close button, see this: https://stackoverflow.com/questions/26828482/adding-a-terminate-playback-button-using-android-l-notification-mediastyle

> When using the Support Library's version of this notification style, i.e. NotificationCompat.MediaStyle, there is the setShowCancelButton() button.
>
> This will add the close button, but only on versions prior to Lollipop, to work around a bug with notifications that cannot be made dismissable after being displayed as part of a foreground service.
>
> **In Lollipop the preferred option is to just being able to dismiss the notification (when the audio is paused) instead of having a custom button to do that.**

In the last commit I have implemented setShowCancelButton() (for older APIs) and made sure that the notification is dismissable when playback is stopped.

tzugen:
Regarding the badge, I have now disabled it in the Notification Channel. But an app cannot modify the channel after it was created, so if you have been testing this before it won't make a difference and you will need to disable the badge in the Android Notification settings.

ultrasonic/src/main/java/org/moire/ultrasonic/service/MediaPlayerService.java
```diff
@@ -665,6 +644,21 @@ public void updateNotification(PlayerState playerState, DownloadFile currentPlay
...
+     * Create a Notification channel on API >= 26
+     */
+    private void createNotificationChannel() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            //The suggested importance of a startForeground service notification is IMPORTANCE_LOW
+            NotificationChannel channel = new NotificationChannel(NOTIFICATION_CHANNEL_ID, NOTIFICATION_CHANNEL_NAME, NotificationManager.IMPORTANCE_LOW);
+            channel.setLightColor(android.R.color.holo_blue_dark);
+            channel.setLockscreenVisibility(Notification.VISIBILITY_PUBLIC);
+            // Don't show a red dot on the icon
+            channel.setShowBadge(false);
```
tzugen: Disabling the badge here

tzugen:
I added a Star/Unstar action :)

nitehu:
I've tested this and something isn't right, if I pause Ultrasonic now Android just kills the service after 1-2 minutes and removes the notification. (Maybe this behavior was what the Poweramp guys worked around with the "x" ?)

tzugen:
Yes, because onPause we call stopForeground(false). Do you think it
would be good to keep the service running?

ogarcia:
> I've tested this and something isn't right, if I pause Ultrasonic now Android just kills the service after 1-2 minutes and removes the notification. (Maybe this behavior was what the Poweramp guys worked around with the "x" ?)

But I think that this is the expected. If the app is paused the notification will be gone with the time.

Let me test if YT Music do the same...

ogarcia:
> Let me test if YT Music do the same...

Nope, YT Music maintains the notification until you swipe it or you close the app.

nitehu:
> But I think that this is the expected. If the app is paused the notification will be gone with the time.

From the user's viewpoint "it just randomly disappears". Surely this can't be the best practice, I'd be annoyed by this as it feels more like a bug. I don't know any other Android apps which would remove notifications randomly. (Also currently there is a real bug with this solution, as Ultrasonic forgets the position in the song when this happens...)

I've checked some random music players from the play store. 8/10 uses the "x" solution, the other 2 removed the notification after a few minutes. 

I know I'm conservative, but I liked how Ultrasonic worked before :)
If it can't be solved to show the notification until it is dismissed manually, I'd vote for the "x".

tzugen:
@ogarcia what do you think the behaivour should be?

ogarcia:
> @ogarcia what do you think the behaivour should be?

I think that the best approach is like YT Music. It creates the notification when you start play music, and, if you pause it, the notification remains until you swipe it or close the app. But I don't know if the notification will be disappear with the time in low memory devices, in my case seems keep active forever.

At this moment I see that the notification is created when you start Ultrasonic instead of when you start play music. I think that is _more elegant_ that the notification is created when you start play music and let it stay there until you swipe it (pausing or stopping music before, of course) or close the app. But in case that keeping the notification with music paused is not possible I still prefer this new approach that the old one with `X` button

nitehu:
Ok for me.
So I think the paused - stopped behavior of Ultrasonic must be reworked then.
- Currently when Ultrasonic is started, it is in the paused state (if it wasn't stopped before), that's why it starts with the notification (and the mini-now-playing) visible. Should it start in stopped state (except of course when started by a play event)?
- When the service is killed, Ultrasonic goes to the stopped state, and the playback position in the song resets. I think the playback position should be saved so the playback can continue where it left off.

What do you guys think?
Or should we create a new issue for this and just fix later?

ogarcia:
> Should it start in stopped state (except of course when started by a play event)?

Agree

> * When the service is killed, Ultrasonic goes to the stopped state, and the playback position in the song resets. I think the playback position should be saved so the playback can continue where it left off.

Here we have a problem. If service is killed because I swipe the notification Ultrasonic must should be in stopped state.

> What do you guys think?
> Or should we create a new issue for this and just fix later?

@tzugen what you prefer? Do you want to try fix this or create a new issue to we fix it later?

nitehu:
> Here we have a problem. If service is killed because I swipe the notification Ultrasonic must should be in stopped state.

I agree, and when the playback service isn't running the state must be stopped anyways. 
The problem is with saving the playback position. I'm thinking about the case when someone plays a song on e.g. a BT headset, they pause the playback for a minute, but next time when they press play the song will be reset to the start if we don't save the position.

Did you mean that when the notification is swiped, the position must be reset (like old stop behavior), but when Android kills it, the position can be saved?

ogarcia:
> Did you mean that when the notification is swiped, the position must be reset (like old stop behavior), but when Android kills it, the position can be saved?

Forgive it. It can store play position in any case, in fact YT Music always store play position.

Maxmystere:
Hello ! Nice PR that I wanted to do for some time but never found the time to anyway looks good but still some missing functionality on Android 11
And as already told swiping notification doesn't stop music so that can make it a bit confusing where the audio can come from

Also Spotify does like it's not possible to swipe notification if it's playing and if app is killed it will stop music and remove notification

I can run tests for you if you desire
Screenshot_20210420-181900__01
Screenshot_20210420-182639


This looks smooth
Screenshot_20210420-181855__01

tzugen:
> Hello ! Nice PR that I wanted to do for some time but never found the time to anyway looks good but still some missing functionality on Android 11

Yes, there is no code yet to support the seekbar feature. It could be added later.

tzugen:
1. Regarding the start behaviour, since this is something outside of the MediaPlayerService class, I would rather look at that separately. I have opened a bug #426 
2. I made sure now that playback is stopped when the notification is dismissed (using `setDeleteIntent`)
3. It tries to detach the notification now when exiting Foreground. This is flag (`STOP_FOREGROUND_DETACH`) is API >=24 only. On lower APIs the notification might still disappear after some time, when the service is killed due to low memory.

I think that's it :sweat_smile:

nitehu:
Something is still fishy... on my phone (Samsung Note 9 w Android 10) the MediaPlayerService is killed after 3-4 minutes **even when I'm playing music** (over BT headset, the phone is turned off) o_O
This morning I tried to play some music, the service was killed and the notification was removed.
After updating with your latest commits now the service is still killed but the notification remains...

There isn't any clue in the logcat:

```
2021-04-21 09:08:59.465 9056-9056/org.moire.ultrasonic.debug I/CancellableTask: Cancelling BufferTask 
2021-04-21 09:08:59.577 9056-9056/org.moire.ultrasonic.debug I/LocalMediaPlayer: STARTED -> IDLE 
2021-04-21 09:08:59.580 9056-12351/org.moire.ultrasonic.debug E/MediaPlayerNative: error (-38, 0)
2021-04-21 09:08:59.613 9056-9056/org.moire.ultrasonic.debug I/CancellableTask: Cancelling BufferTask 
2021-04-21 09:08:59.617 9056-9056/org.moire.ultrasonic.debug I/LocalMediaPlayer: LocalMediaPlayer destroyed
2021-04-21 09:08:59.623 9056-9056/org.moire.ultrasonic.debug I/Downloader: Downloader stopped
2021-04-21 09:08:59.628 9056-9056/org.moire.ultrasonic.debug I/ShufflePlayBuffer: ShufflePlayBuffer destroyed
2021-04-21 09:08:59.638 9056-9056/org.moire.ultrasonic.debug I/MediaPlayerService: MediaPlayerService stopped
2021-04-21 09:08:59.644 9056-9056/org.moire.ultrasonic.debug W/MediaPlayer: mediaplayer went away with unhandled events

```

Edit: Some more info.
I've added a log to see if the service is in the foreground (by ActivityManager.getRunningServices)
It seems when I play one long track, the service isn't killed, and it stays in the foreground.
When Ultrasonic changes to the next track, after it updates the notification and calls startForeground, the service won't be in the foreground as reported by the ActivityManager...
I'm quite baffled by this, will continue testing when I have more time...

tzugen:
I think I fixed it. Can you try again? @nitehu

nitehu:
No luck unfortunately...
It's better with this fix as the service doesn't go to background at every track change.
However, when I press pause and then play again, it won't come out to the foreground, and will still get killed.
Please see the logs attached.

[log.txt](https://github.com/ultrasonic/ultrasonic/files/6351330/log.txt)

I'm using [this code](https://stackoverflow.com/a/36127260/4207238) to check if it is in the foreground or not. 

Edit:
A clue: when the service says it's not in foreground after the startForeground(), the notification doesn't gets updated anymore (e.g. the play button stays even if it should change to pause). 
This only seems to happen while Ultrasonic is in the background, controlled from the notification.

tzugen:
This was getting to complex. I removed the modifications to stopForeground and simply added a close button. So the behaviour when and how the service is started or stopped should be exactly the same.  I think in a separate bug we can discuss if or how to change it. I can only say that the logic is very brittle ;)

nitehu:
>  I think in a separate bug we can discuss if or how to change it. I can only say that the logic is very brittle ;)

I agree, that's a good idea. I'm starting to think it has something to do with my phone's OS. I can't reproduce this behavior in any emulators...

With the latest commits it seems to work all right, no sudden kills (at least for now, I'll test it further).

Also in the meantime I've found two other issues:
- Bluetooth headset buttons aren't working during playback
- The lock screen media controls are missing

Maybe these are related, as I remember Android handles them similarly.

tzugen:
The Bluetooth i have to try out.
The lockscreen puzzles me. I have the same API version as you and I see the lockscreen controls.

21.04.2021 18:11:53 Nite ***@***.***>:

>> I think in a separate bug we can discuss if or how to change it. I can only say that the logic is very brittle ;)
>> 
> I agree, that's a good idea. I'm starting to think it has something to do with my phone's OS. I can't reproduce this behavior in any emulators...
> 
> With the latest commits it seems to work all right, no sudden kills (at least for now, I'll test it further).
> 
> Also in the meantime I've found two other issues:
> 
* > Bluetooth headset buttons aren't working during playback
* > The lock screen media controls are missing

tzugen:
Bluetooth is fixed.
Regarding lockscreen:

> In Android 5.0 (API level 21) or greater the system does not provide transport controls on the lock screen. Instead, you should use a MediaStyle notification to display transport controls.

So it seems this is the intended behaviour. But you don't see the notification in the lockscreen at all? Here it shows up correctly.

nitehu:
> So it seems this is the intended behaviour. But you don't see the notification in the lockscreen at all? Here it shows up correctly.

Argh this too may be another Samsung problem after all...
So until now I've used small notification icons only on my lock screen, and the media controls were above the notifications. Also the Always-On-Display showed the actually playing song (I can't screenshot that...)

Screenshot_20210421-201859_Ultrasonic

Now I enabled to show the full notifications, and this way my lock screen looks similarly to yours:

Screenshot_20210421-202728_Ultrasonic

I can replace the media player widget with a clock so that place wont look so empty. Unfortunately there isn't a solution for the Always-On-Display, maybe Samsung uses the deprecated API for that... (Not that I was using these features a lot, just noticed that they are missing)

Is this how it is intended to work with the MediaSession?
@ogarcia @Maxmystere could you guys please try out the lock screen on your phones when you have a few free minutes and give some feedback on what you think?

nitehu:
I've checked with an old Nexus S (API19). This was before the changes:



After the changes the lock screen controls are gone.



But isn't MediaSessionCompat should be backward compatible?

Maxmystere:
On my oneplus with android 11 it looks like this
https://developer.android.com/guide/topics/media/media-controls
Screenshot_20210421-212145
Screenshot_20210421-212340

tzugen:
I guess if i resurrect the RemoteController stuff the lockscreen controls in older devices will come back. Don't know if it's worth it, but also I dont have that problem.

tzugen:
Here is more info on it:
https://stackoverflow.com/questions/30942054/media-session-compat-not-showing-lockscreen-controls-on-pre-lollipop

nitehu:
I'm afraid we are definitely missing something here.
I gave youtube music trial a go, and this is how it looks like on my lock screen:

Screenshot_20210422-072140_YouTube Music

When unlocked, it uses the new notification, but on the lock screen the notification is not visible, it displays the lock screen media controls (and also Always-On-Display works). Does Google use their own deprecated API to achieve this?
I'd say we should give this feature another few days of googleing to see if maybe this behavior can be reproduced somehow.

tzugen:
Ok, I might have gotten to the root of it. Unfortunately I can't test it because emulator at API 19 refuses to connect to my https server. Can you check it out @nitehu ?

nitehu:
Wow this is really going somewhere! Thanks very much for your patience.

So now the lock screen controls appeared on both my Note and Nexus S. There are still minor problems with it:
- the control keys are disabled / missing. This is because a `playbackState.setActions()` call is missing it seems
- The song details are displayed correctly, but the button isn't updated when play/pause is pressed. Interestingly enough when I added a breakpoint before playbackState.setState(), the buttons sometimes changed... so it may be related to concurrency... I haven't found what causes this yet.

tzugen:
Ok, now one thing lead to another.. A simple MediaPlayerService translation to Kotlin then required to get detekt up-to-date...
But now the Actions are set. For now only Play, Pause and Stop, because these are the methods currently implemented in the Session. If that works I can implement skip as well.

Regarding this:
>The song details are displayed correctly, but the button isn't updated when play/pause is pressed. Interestingly enough when I added a breakpoint before playbackState.setState(), the buttons sometimes changed... so it may be related to concurrency... I haven't found what causes this yet.

Which button where? In the lockscreen or in the notification?

nitehu:
> Which button where? In the lockscreen or in the notification?

Oh sorry, this is the play/pause button on the lockscreen. For me it always shows pause, but otherwise it works.
Also I've just noticed that when I dismiss the notification with the "x", the lockscreen controls stay visible. They only disappear when I exit Ultrasonic. So something is wrong with updating it.

The notification itself is working fine for me it seems. It looks good on my old Nexus S too.

tzugen:
You are dismissing the notification with the X button that I added with .addiction() ?

nitehu:
Yep. The notification goes away, the lock screen controls stay as they
were...

tzugen:
@nitehu  I added some debug log calls. Can you run the latest commit, close the notification with the X and post the log of everthing that happened then?

nitehu:
Argh now that fixed it! Maybe it was only needed to set the actions properly... Now the lock screen buttons work correctly for me, the play changes to pause when playing and changes back when paused.

Also I've tried out what if I added the `ACTION_SKIP_TO_NEXT `and `ACTION_SKIP_TO_PREVIOUS`, because the buttons were grayed out on the lock screen. After I added these two, now the next and previous buttons are enabled and working correctly.
This means all the lock screen controls are working as intended, thank you for your help and efforts.

I've tried out the previous bug, where after closing the notification with the "x" the lock screen controls stay visible. This behavior remains, but now the lock screen is in stopped state, so its not that annoying.
Anyways, here is the log, please check it for any strangeness:

[log1.txt](https://github.com/ultrasonic/ultrasonic/files/6363486/log1.txt)

There is still an issue (I hope this is the last one regarding the lock screen :)
So after creating the previous log, I've restarted Ultrasonic from the lock screen by pressing the still visible play button, and this happened:

[log2.txt](https://github.com/ultrasonic/ultrasonic/files/6363506/log2.txt)

Ultrasonic started playback as normal, with a nice notification. But when now I pressed the "x" on the notification, the playback stopped but the notification remained in place. It won't disappear even if I exit Ultrasonic :)
There is only one way to dismiss this stuck notification: If I press play again on the stuck notification, Ultrasonic seems to recreate it (it blinks for a second), and now the "x" works again on it.
I've created a log of this too:

[log3.txt](https://github.com/ultrasonic/ultrasonic/files/6363563/log3.txt)

I've checked with Poweramp, after pressing the "x" in its notification the lock screen controls are gone. I think if we could achieve the same somehow, the whole "stuck notification" would be solved too.

But this is such a minor inconvenience now, maybe affecting a few users with similar phones, and I see that you can't reproduce it on your end. If you can't find what causes it, maybe we should drop the lock screen tests, and write this up in a new issue as a bug. It may well be that this will fix itself later as we move to newer APIs anyway.
(On the other hand there is a possibility that this behavior is a signal of some deeper problems which may just cause pain later if we don't fix it now...)
Let me know what you think.

tzugen:
Hi :)

I added a commit to map the PlayerState more precisely. If it is STOPPED then also the PlaybackCompatState ist stopped, and maybe then the lockscreen controls should disappear as well.

Regarding the stuck notification, I am pretty sure the bug is in the logic of `showWhenPause` and this was the same place I had touched before because it was quite messy, but that had all those unintended side effects with the lifecycle of the service. 
If you don't mind, maybe you could add this case to the bug I already opened about this.

tzugen:
Hi :)

I added a commit to map the PlayerState more precisely. If it is STOPPED then also the PlaybackCompatState ist stopped, and maybe then the lockscreen controls should disappear as well. 

Regarding the stuck notification, I am pretty sure the bug is in the logic of showWhenPause and this was the same place I had touched before because it was quite messy, but that had all those unintended side effects with the lifecycle of the service.
If you don't mind, maybe you could add this case to the bug I already opened about this.

tzugen:
I think in general if it is just minor minor things, it would for me make more sense to finish up anything that blocks merging,
and then later spent the time to implement more of the modern Android API. A lot of these small hickups we are seeing is probably because we are just implementing some of those APIs, and if we implement more, than for example a androidx.media2.player will generate the PlayerState itself, and we won't have to worry about setting the right values anymore.

nitehu:
I've tested the newest version but the behavior is the same.

> A lot of these small hickups we are seeing is probably because we are just implementing some of those APIs, and if we implement more, than for example a androidx.media2.player will generate the PlayerState itself, and we won't have to worry about setting the right values anymore.

I agree, while I was browsing other open source players' code they usually don't handle the PlayerState manually. So it will hopefully solve itself.

I'll create an issue about this for now, and test everything else on the weekend.

ogarcia:
@nitehu seeing the great work that @tzugen is doing, I'm going to give him a member status in Ultrasonic organization. @tzugen this means that you will can organize bugs and aprobe pull requests, but the workflow is the same, every change must be in its PR and must be reviewed by all of us :wink:

nitehu (approved):

nitehu:
Ok the code seems to be fine. I also haven't found any new issues, everything seems to be working.
@tzugen could you please change the new warning logs you've added back to debug level?
And please let us know if you are planning to change/add anything else or if this can be merged.

tzugen:
@ogarcia  Thank you, I am looking forward to continue with all of you :)

@nitehu Nice! I just pushed a commit to adjust the timber levels. I don't plan any further changes here.

ogarcia:
I think that is all fine. I'll do tomorrow more tests and will merge.

What do you think? New release or you want add something more?

tzugen:
I think it would be good to release and not add more feature to this one (2.20)
Do we uses the beta track on Play Store? It would be also an option to release it first as beta.
Or do you have an established procedure there?

ogarcia:
> I think it would be good to release and not add more feature to this one (2.20)

Ok. Fine :+1: 

> Do we uses the beta track on Play Store? It would be also an option to release it first as beta.
> Or do you have an established procedure there?

Yes. The procedure is that the latest version is always in beta channel (for example in this case 2.20.0). If the version has issues then we will fix releasing a 2.20.1 in beta channel too.

When we will publish the 2.21.0 the 2.20.x will be moved to stable channel.

In F-Droid there is not beta channel so latest version is always published when we make a new tag.

tzugen:
@ogarcia  Can this be merged? (I don't want to rebase it to many times onto develop...)

ogarcia:
@tzugen yes, merge it. I cannot release new version yesterday because my hard disk die :cry:, but no worry I have backups :wink:

tzugen:
Oh no! I hope the backups work!

I cannot merge... I think I am still missing a permission...

tzugen:
🥳

PR #424, opened 2021-04-18
Remove gradle warning

tzugen:
Just a minor fix. Removes this:
```
WARNING: API 'variant.getGenerateBuildConfig()' is obsolete and has been replaced with 'variant.getGenerateBuildConfigProvider()'.
It will be removed in version 5.0 of the Android Gradle plugin.
```


nitehu:
Thanks for fixing this!

nitehu (approved):

nitehu:
Eh I've just noticed that the okHttp 3.14.9 requires API level 21. Currently Ultrasonic runs on 14.
@tzugen Was there a specific reason to update it, or should we return to use 3.10.0?

tzugen:
Oh, no! I opened a PR with a revert..

PR #431, opened 2021-04-26
Add a retry count to stop infinite downloads

### LINKED ISSUE #416: Pinning can result in download loop
## Problem description

The other day I pinned a bunch of songs and left the house. In the afternoon I got a warning from my data provider that I had chewed up 9GB of data that day! Turns out ultrasonic was continually downloading one of the songs over and over and over and over.

### Steps to reproduce

Go to album view, select the check box next to a particular song, hit the little pin icon at the bottom. 

## System information

 * **Ultrasonic version**: 2.18.0 (and also develop branch)
 * **Android version**: 11 (and also emulator)
 * **Device info**: Pixel 3
 
## Additional notes

This one scares me a bit, so I think I'll take it on as my next bug fix.

SaintDubious:
See also: https://github.com/ultrasonic/ultrasonic/issues/416

The `Downloader` class keeps a list of background files to download. A separate thread will check on the downloads every five seconds. When a download fails for some reason it will be re-attempted. Unfortunately there is no cutoff, so a file will be re-downloaded forever. This can have a serious impact on monthly bandwidth charges.

This fix introduces a simple `retryCount` to the `DownloadFile` class. On each failure the count will be decremented. If the count gets to zero then a new `shouldRetry()` function will return false. The `Downloader` class will check for this condition and stop attempting to download if it is hit.

While this fix does work I worry that it may be overly simplistic. I am not familiar with the entire download mechanism and I see that it's used very widely throughout the system. My hope is that one of the reviewers will have more familiarity and can spot potential issues. The last thing I want to do is add another edge case.

### DIFF
--- a/ultrasonic/src/main/java/org/moire/ultrasonic/service/Downloader.java
+++ b/ultrasonic/src/main/java/org/moire/ultrasonic/service/Downloader.java
@@ -194,6 +194,12 @@ public class Downloader
                     revision++;
                     i--;
                 }
+                else if (downloadFile.isFailed() && !downloadFile.shouldRetry()) {
+                    // Don't continue to attempt to download forever
+                    backgroundDownloadList.remove(i);
+                    revision++;
+                    i--;
+                }
                 else
                 {
                     currentDownloading = downloadFile;
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/DownloadFile.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/DownloadFile.kt
@@ -45,6 +45,7 @@ class DownloadFile(
     private val mediaStoreService: MediaStoreService
     private var downloadTask: CancellableTask? = null
     var isFailed = false
+    private var retryCount = 5
 
     private val desiredBitRate: Int = Util.getMaxBitRate(context)
 
@@ -125,6 +126,10 @@ class DownloadFile(
 
     fun shouldSave(): Boolean {
         return save
+    }
+
+    fun shouldRetry(): Boolean {
+        return (retryCount > 0)
     }
 
     fun delete() {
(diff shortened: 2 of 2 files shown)


tzugen:
The code looks ok to me. I haven't tested it. 

Kind of offtopic: 
I think the class could be refactored more radically. Right now it takes care of the 1. the playlist and 2. the download queue.
IMO it would be good to separate the logic between these two things. The androidx.media2.MediaPlayer comes with a build-in playlist. If we switch to that, we can free the Downloader Class of its "dual responisbilities" :) 
What I think could be also nice is to implement a `PriorityBlockingQueue` instead of our own Queue managment.

But this is just some ideas I had regarding this class, I don't mean it in anyway as an obligation to you to implement them ;)

tzugen:
The code looks ok. I haven't tested it yet.

Considering this can lead to extra costs for users, might warrant it to include it in 2.20?

What do the others think? @ogarcia @nitehu

ogarcia:
I cannot test it since I don't have Android 11 but change look OK

nitehu (approved):

nitehu:
I've tested this with an emulator. The download loop stops after 5 retries as expected.
This fix only works for songs that are in the backgroundDownloadList (means those that were explicitly selected to be downloaded / pinned by the user). 
Just a note - songs on the playlist can still get stuck in an endless loop, which is a separate issue (and an UI challenge, as I think those can't just simply be removed...)

@SaintDubious should we close #416 with this?

SaintDubious:
@nitehu yeah probably cool to close https://github.com/ultrasonic/ultrasonic/issues/416 . There's still the issue of what actually causes my downloads to fail, but those can be worked out as minor bug fixes as time allows. I think this handles the more potentially expensive problem.

Sorry for the delayed response, I can only work after I've put the kids to bed, after 9:30pm EST.

nitehu:
No worries, that's perfectly fine :)

PR #436, opened 2021-04-27
Update Room from 2.2.6 to 2.3.0

tzugen:
This gets rid of the warning `Incremental annotation processing requested, but support is disabled because the following 
androidx.room.RoomProcessor (NON_INCREMENTAL).`

Room has enabled incremental support in 2.3.0. For us this means slightly faster build times.


tzugen:
This is for 2.21.X

tzugen:
Will do this later

PR #445, opened 2021-04-27
Implement a room database for better offline support

tzugen:
In core/domain we have nice Kotlin data classes, which describe the contents of the responses we can get from the API (eg. list of Artists, Albums etc.)

These can be easily annotated with Room `@Entitys`. That makes them storable in the room database. 

I propose to annotate most of the domain classes this way and to add the respetive DAOs.

This way we can reduce the number of current storage solutions (3: room, timedcache, permanentfilestorage) to one (room).

In a further step, we can then modify the OfflineMusicService class to use data from the database if available. 

Cache invalidation (like always) will require some thought of course.


ultrasonic/src/main/java/org/moire/ultrasonic/service/CachedMusicService.kt
```diff
@@ -52,18 +52,21 @@ class CachedMusicService(private val musicService: MusicService) : MusicService
...
-    private val cachedArtists = TimeLimitedCache<Indexes>(60 * 60, TimeUnit.SECONDS)
     private val cachedPlaylists = TimeLimitedCache<List<Playlist>?>(3600, TimeUnit.SECONDS)
     private val cachedPodcastsChannels = TimeLimitedCache<List<PodcastsChannel>?>(3600, TimeUnit.SECONDS)
     private val cachedMusicFolders = TimeLimitedCache<List<MusicFolder>>(10 * 3600, TimeUnit.SECONDS)
     private val cachedGenres = TimeLimitedCache<List<Genre>>(10 * 3600, TimeUnit.SECONDS)
 
+    // New Room Database
+    private val cachedArtists =  appDatabase.artistsDao()
+
     private var restUrl: String? = null
```
tzugen: Because I used the same method calls in the Room DAO as in the TimeLimitedCache, changing the cache provide is a one liner..

nitehu:
Actually I'm not that familiar with this part of Ultrasonic, but this work seems very promising to me.

- For now I'm not entirely sure which cached objects can we use in the Offline mode to provide useful features to the users. As Ultrasonic is mainly a music player, the most important offline features are dependent on the downloaded songs. So e.g. the recent lists cached from the server won't mean a lot in offline mode, if the songs aren't available offline. But this doesn't mean that we won't find useful offline features based on this cache in the future.
- I absolutely agree to refactor these caches and other data storage to use Room. Especially if it removes legacy code :) It seems to be a lot clearer and easier instead of using different kinds of serialization/deserialization.
- Invalidating the caches maybe can be done based on time for now (as it were before), and later this can be fine-tuned if necessary.
- I think this work will have huge positive impact on storing and handling the playlist later. Instead of loading and saving it constantly we can use more modern ViewModels and LiveData.
- Just a side note, I'm not sure what is the performance of Room, but it isn't unusual for Ultrasonic power users to have thousands of entries in their library / playlist, so we must keep that in mind.

Also we should ask @ogarcia for opinions :)

ogarcia:
Agree with @nitehu but I think that some information can be useless, for example being able to show _Recent Albums_ in offline mode has not much sense if you cannot play it. Anyway I think that this can be achieved invalidating the cache so implement this change is a great idea.

tzugen:
I believe this is ready now:

* For now the Cache is only used for Artists, Indexes and Music Folders
* There is one database for each server
* Index items are saved with a "musicFolderId" prop, which makes it possible to filter the database by musicFolder without necessarily having to query the server for it.
* Databases for new Servers are created on the fly
* If the user removes a server, the respective database is deleted.

In the future this functionality could be added:
* When a user downloads an album, it's entry could be stored in an offline database. Then if the user wants to list all albums while offline, we can use the database to provide a nice view of them.

nitehu (approved):
Tested this, didn't find any problems.

tzugen:
(Rebased on dev branch)

Issue #426, opened 2021-04-20
Start service only when playback is started by the user


tzugen:
>At this moment I see that the notification is created when you start Ultrasonic instead of when you start play music. I think that is more elegant that the notification is created when you start play music and let it stay there until you swipe it (pausing or stopping music before, of course) or close the app. But in case that keeping the notification with music paused is not possible I still prefer this new approach that the old one with X button

tzugen:
Fixed by #726

PR #447, opened 2021-05-01
Get rid of rogue session

tzugen:
Using [Media Controller Tester](https://developer.android.com/guide/topics/media-apps/audio-app/media-controller-test) I finally figured out that there was an empty media session being created on start.

It took quite some time to figure out where it was coming from, but now it should be gone.

@nitehu Can you test this?

### DIFF
--- a/ultrasonic/src/main/java/org/moire/ultrasonic/receiver/BluetoothIntentReceiver.java
+++ b/ultrasonic/src/main/java/org/moire/ultrasonic/receiver/BluetoothIntentReceiver.java
@@ -29,7 +29,7 @@ import org.moire.ultrasonic.util.Constants;
 import org.moire.ultrasonic.util.Util;
 
 /**
- * Request media button focus when connected to Bluetooth A2DP.
+ * Resume or pause playback on Bluetooth A2DP connect/disconnect.
  *
  * @author Sindre Mehus
  */
@@ -63,7 +63,6 @@ public class BluetoothIntentReceiver extends BroadcastReceiver
 		if (state == android.bluetooth.BluetoothA2dp.STATE_CONNECTED) actionA2dpConnected = true;
 		else if (state == android.bluetooth.BluetoothA2dp.STATE_DISCONNECTED) actionA2dpDisconnected = true;
 
-		boolean connected = actionA2dpConnected || actionBluetoothDeviceConnected;
 		boolean resume = false;
 		boolean pause = false;
 
@@ -81,12 +80,6 @@ public class BluetoothIntentReceiver extends BroadcastReceiver
 				break;
 			case Constants.PREFERENCE_VALUE_A2DP: pause = actionA2dpDisconnected;
 				break;
-		}
-
-		if (connected)
-		{
-			Timber.i("Connected to Bluetooth device %s address %s, requesting media button focus.", name, address);
-			Util.registerMediaButtonEventReceiver(context, false);
 		}
 
 		if (resume)
--- a/ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
+++ b/ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
@@ -679,7 +679,7 @@ public class Util
 
 	public static void registerMediaButtonEventReceiver(Context context, boolean isService)
 	{
-		if (getMediaButtonsPreference(context))
+		if (getMediaButtonsEnabled(context))
 		{
 			if (isService) mediaButtonsRegisteredForService = true;
 			else mediaButtonsRegisteredForUI = true;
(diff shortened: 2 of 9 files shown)


nitehu (approved):

nitehu:
Something isn't right, a fresh install on a Pixel XL API30 emulator starts with an empty notification visible now.
Screenshot 2021-05-02 124942
I haven't even added servers yet...
It seems that it changes to the correct new MediaStyle notification when I start to play music. When I press the "x" it disappears.
But then when I exit ultrasonic (using the menu), the empty notification comes back!

tzugen:
And this empty notification is not appearing on the "develop" branch?

02.05.2021 12:58:26 Nite ***@***.***>:

nitehu:
Hmm you made me worry for a second there, but I've tested it again now :) 
The develop branch works correctly, the notification only appears when there is a playlist loaded and Ultrasonic is in the play/paused state. 
On the rogue-session branch however an empty notification appears after startup and exit too...

tzugen:
OK good to hear!

But confusing that this patch changes the behavior of the notification...

nitehu:
It may be because in the Utils the (un)registerMediaButtonEventReceiver calls executeOnStartedMediaPlayerService, so that starts the service (and that displays the notification)
The (un)registerMediaButtonEventReceiver is called from the NavigationActivity at the start/stop of Ultrasonic.

I think originally it was OK to register the event receiver at the start, but now that it requires a MediaSession this must be changed too. I'm just not sure how will Ultrasonic receive button presses before it creates a MediaSession (e.g. pressing play on a headset while Ultrasonic isn't running) but the new API may have a different solution to that.

tzugen:
We should try to define first in which states or conditions the service is running, under which conditions the service is foregrounded etc.

Then it will become clear what to do in code. (Also regarding #426 )

I made a start here: https://github.com/ultrasonic/ultrasonic/wiki/Playback-states-&-service-states

Please feel free to change it, especially in the columns where I was not sure (`?`)

nitehu:
> Please feel free to change it, especially in the columns where I was not sure (?)  

I've filled the ?s and added some notes.

Edit:
[From this](https://developer.android.com/guide/topics/media-apps/mediabuttons) it seems we do not need to register for the buttons when there is no MediaSession.
Ultrasonic will receive buttons in the foreground Activity, or as an Intent...

tzugen:
That is great, thanks a lot for finding this :+1: 

I removed all calls from the UI to the Util function and that simplifies things a lot, and the media buttons work correctly here.
Even when I force quit the app etc.

ultrasonic/src/main/java/org/moire/ultrasonic/receiver/BluetoothIntentReceiver.java
```diff
@@ -83,12 +82,6 @@ else if (BluetoothDevice.ACTION_ACL_DISCONNECTED.equals(action) || BluetoothDevi
 				break;
 		}
 
-		if (connected)
-		{
-			Timber.i("Connected to Bluetooth device %s address %s, requesting media button focus.", name, address);
-			Util.registerMediaButtonEventReceiver(context, false);
-		}
-
```
tzugen, line 91: This was actually a bug, because it didn't respect the preference!

ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
```diff
@@ -677,41 +677,16 @@ public static Bitmap scaleBitmap(Bitmap bitmap, int size)
...
-					{
-						mediaPlayerService.registerMediaButtonEventReceiver();
-						return null;
-					}
-			);
-
+		MediaPlayerService mediaPlayerService = MediaPlayerService.getRunningInstance();
+		if (mediaPlayerService != null) {
+			mediaPlayerService.updateMediaButtonReceiver();
 		}
```
tzugen, line 687: Only execute this call, if we have a running service...

ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
```diff
@@ -677,41 +677,16 @@ public static Bitmap scaleBitmap(Bitmap bitmap, int size)
...
-		MediaPlayerService.executeOnStartedMediaPlayerService(context, (mediaPlayerService) ->
-				{
-					mediaPlayerService.unregisterMediaButtonEventReceiver();
-					return null;
-				}
-		);
-
-		Timber.i("MediaButtonEventReceiver unregistered.");
-	}
-
```
tzugen: Not needed anymore, because the preference logic is handled inside the service

nitehu (approved):

nitehu:
Nice job!
This works very well for me now on emulator and on a real phone too.
- The media buttons are working properly: during playback, when playback is stopped, and even when Ultrasonic isn't running
- The empty notification is gone
- The problems I had with the lock screen are mostly gone now. It is displayed and removed correctly.

I think for now this can be merged, if we'll find anything new a new issue can be created for it.

mase76:
I also have this empty session with 2.23.1. Every time after rebooting the device, and sporadic  from time to time.

nitehu:
@mase76 could you please give us some more details about the empty session? Where do you see it? Is it an empty notification? Do you have a playlist loaded? What do you do to make it go away?
Any info would be nice.

mase76:
It is an empty notification like nitehu showed above. I have no playlist loaded. I hit the X to move it away.

mase76:
With 2.24.0 I still get this empty notification. However there is no more X to close it. I have to start Ultrasonic and kill the application to get rid of the notification. No playlist is loaded. The notification appears always on boot and also from time to time.

nitehu:
Created #621 for this problem.

PR #451, opened 2021-05-01
Remove MediaStoreService, reinstate Util.scanMedia

### LINKED ISSUE #434: Albumart id can be negative?
> If the album_id is a negative number (for example a URI like "content://media/external_primary/audio/albumart/-824043666") then the query above on line 111 actually throws an exception with a message: "java.lang.UnsupportedOperationException: Unknown or unsupported URL: content://media/external_primary/audio/albumart/-824043666"

From: https://github.com/ultrasonic/ultrasonic/pull/414#commitcomment-50040677

I would have assumed that id are always positive and an id of -1 (or -2 for that matter) denotates an invalid id.

@SaintDubious Can you find out if it is your server which returns this negative id?

### LINKED ISSUE #435: Android MediaStore API is broken in API >= 29
There used to be two ways to add music to the Android Media Library.

1. Insert the data to the MediaStore with a reference to the Uri of the file in `MediaStore.MediaColumns.DATA`
2. Trigger a scan with `ACTION_MEDIA_SCANNER_SCAN_FILE`

Both of them have been deprecated with API 29.  And the deprecation in `ACTION_MEDIA_SCANNER_SCAN_FILE` tells to use the MediaStore. Lol....

So there seems to be no supported way to insert media files stored in our data directory to the Media Library. Maybe this is because Google is pushing `MediaBrowserService`?

Anyway, speculation aside, what shall we do? 

I think until we can get to implementing `MediaBrowserService`, which will be a bunch of work, it seems the only sensible thing is to stop inserting stuff into MediaStore on API >=29

tzugen:
Because MediaStore never officially supported adding external media using `DATA`, and now on >=29 it doesn't work at all anymore and causes a number of issues, I am removing it completely. See also the discussion in #435 

This brings back the Util.scanMedia method. `Intent.ACTION_MEDIA_SCANNER_SCAN_FILE` is also deprecated, but it is only a broadcast, so we call safely call it on all APIs, and if the platform has a fitting BroadcastReceiver registered, it will scan the media, otherwise the broadcast will end up in Nirvana (or someplace similar).

Sorry for the forth and back!


ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/DownloadFile.kt
```diff
@@ -137,7 +135,8 @@ class DownloadFile(
         Util.delete(partialFile)
         Util.delete(completeFile)
         Util.delete(saveFile)
-        mediaStoreService.deleteFromMediaStore(this)
+
+        Util.scanMedia(context, saveFile)
```
tzugen, line 139: This might look funny, but is ok, because the Uri of the deleted file is still stored in the File Class

nitehu:
I think this is OK for now.
Some filesystem access API changes in Android are around the corner, so sooner or later we must rethink how Ultrasonic stores the media files anyways.

PR #452, opened 2021-05-01
Create dependabot.yml

tzugen:
For more info see the discussion :
https://github.com/ultrasonic/ultrasonic/discussions/449


tzugen:
target_branch: develop is missing! 

And should we do it maybe rather weekly for starters?

ogarcia:
@tzugen I think that is better weekly

tzugen:
Ok, I set it to weekly and configured the target_branch.
I think we can try it out :)

nitehu:
Wow, it already wants to update things...
What do you guys think should be the procedure here? For example the gradle update has failed.

PR #468, opened 2021-05-06
Remove codecov

tzugen:
Just noticed that the codecov bash script was still being executed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,6 @@ jobs:
           command: |
             ./gradlew ciTest testDebugUnitTest
             ./gradlew jacocoFullReport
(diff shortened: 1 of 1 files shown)


ogarcia:
> Just noticed that the codecov bash script was still being executed

:scream: 

Maybe we need update the [android circle ci image](https://circleci.com/docs/2.0/circleci-images/#android) to api-30?

tzugen:
Wait, but then if we have happily executed their bash script all along, we can just reenable CodeCov, no?
Because they have fixed the issue in the mean time.

ogarcia:
The script does not working because the API keys were deleted. Anyway is better disabled.

PR #469, opened 2021-05-07
Add Ampache as supported server

nitehu:
I'd like to add Ampache as a supported and tested server to the readme.
I use it for a year now and Ultrasonic works very well with it. Also their lead, lachlan is usually a great help to us providing issues and fixing Ampache when we find something.


ogarcia (approved):

gerroon:
You can use Ampache as Subsonic mode in Ultrasonic.

cquike:
Hi,
is this available in an official release? I cannot find the option to setup an Ampache account in the current version from F-droid
Thanks!

nitehu:
Ampache is Subsonic API compatible, as well as Ultrasonic. So if you set up / enable the Subsonic API in Ampache, you can add it as any other Subsonic server in Ultrasonic.

PR #476, opened 2021-05-08
Try to remove jcenter()

tzugen:
>**JCenter service update**
>JFrog, the company that maintains the JCenter artifact repository used by many Android projects, made JCenter a read-only repository on March 31st, 2021. According to the announcement, JCenter will allow downloads of existing artifacts indefinitely.
>Developers who publish artifacts on JCenter should migrate their packages to a new host, such as Maven Central.
>Developers who use dependencies from JCenter will need to find the new location of updated versions of those dependencies.


tzugen:
We need to wait for next detekt version

nitehu (approved):

PR #477, opened 2021-05-09
Make all preference operations context-free

tzugen:
Passing around, and especially storing references to context can lead to memory leaks.

Furthermore, I am working to implement the [Android X Paging Library](https://developer.android.com/topic/libraries/architecture/paging/v3-overview) to get endless scrolling in our album views. For this I need API calls which don't require context.

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/RESTMusicService.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/RESTMusicService.kt
@@ -219,7 +219,7 @@ open class RESTMusicService(
         return try {
             if (
                 !isOffline(context) &&
-                Util.getShouldUseId3Tags(context)
+                Util.getShouldUseId3Tags()
             ) search3(criteria)
             else search2(criteria)
         } catch (ignored: ApiNotSupportedException) {
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/RESTMusicService.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/RESTMusicService.kt
@@ -108,10 +108,9 @@ open class RESTMusicService(
 
     @Throws(Exception::class)
     override fun getIndexes(
-        musicFolderId: String?,
-        refresh: Boolean,
-        context: Context
-    ): Indexes {
+        musicFolderId: String,
+        refresh: Boolean
+    ): Indexes? {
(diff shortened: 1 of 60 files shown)


ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/RESTMusicService.kt
```diff
@@ -109,10 +108,9 @@ open class RESTMusicService(
 
     @Throws(Exception::class)
     override fun getIndexes(
-        musicFolderId: String?,
-        refresh: Boolean,
-        context: Context
-    ): Indexes {
+        musicFolderId: String,
```
nitehu: It seems musicFolderId can be null when it isn't selected. Currently this throws this exception for me:

```
W/CommunicationErrorHandler$Companion: java.lang.NullPointerException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkNotNullParameter, parameter musicFolderId
        at org.moire.ultrasonic.service.RESTMusicService.getIndexes(Unknown Source:6)
        at org.moire.ultrasonic.service.CachedMusicService.getIndexes(CachedMusicService.java:139)
        at org.moire.ultrasonic.fragment.ArtistListModel$loadFromServer$2.invokeSuspend(ArtistListModel.kt:94)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

nitehu (changes requested):
Nice work! The code is a lot clearer now again.
However, you've removed nullable ("?") from a few parameters in RESTMusicService for functions that can be called with null from java. This currently causes an exception for me.
Please check these parameters again.

tzugen:
Seems like the IDE has a bug in Refactor>Change Signature, when the interface is in Java, and the implementation is in Kotlin,
then it derives the nullability wrong or something.

I fixed it :+1:

nitehu (approved):
Looking good now!

PR #485, opened 2021-05-11
Removing remaining context from MusicService signatures.

tzugen:
Now there is no mention of Context in MusicService Interface anymore :)


nitehu (approved):
Everything looks and works fine 👍

PR #500, opened 2021-05-19
Fix a lot of warnings, update baseline files, improve CI run times

tzugen:
* Removes our local copy of com.mobeta.android.dslv and instead imports it from maven
* Run detekt in default mode. Type resolution mode is still to experimental and increases build times significantly (thereby also CO2 output...)
* Move a lot of issues suppressed in the detekt baseline to Suppress() tags. This make refactoring easier, because the baseline file doesn't break anymore when someone renames a class.
* Regenerate the lint baseline
* Make lint treat all warnings as errors. Previously lint would return success, even if the commit added new warnings. Now the lint will fail on new warnings, and the CI will stop. This will help us not to introduce new issues.


tzugen:
Run times before: Up to 12 minutes
Runtime now: 8 minutes

nitehu:
Nice job!
I've tested it on an emulator, and unfortunately there seems to be an issue with the DragSortListView. Instead of scrolling in the ListView, it always drags an item and moves it to a new place. It should only do this when the drag handle of the item is moved...
At first glance I'm not sure why it works this way now, the DSLV seems to be set up correctly. I'm a bit afraid Ultrasonic had something modified in the included code...

tzugen:
This was the problem: The "asia.ivity.android:drag-sort-listview" fork which I pulled from Maven was compiled with a newer version of Gradle than the original code, and hit this issue that Gradle warned often about: "Resource IDs will be non-final in Android Gradle Plugin version 5.0". Therefore it failed to get the correct configuration attributes from current_playlist.xml . I found a recently maintained fork https://github.com/masf7g/drag-sort-listview that I used as a base now.

nitehu (approved):
Cool, now everything works

PR #501, opened 2021-05-22
Migrate MediaPlayerController to Kotlin

tzugen:
Also removes context from a number of other classes


tzugen:
This and also the next one are preparations for the database changes.

nitehu (approved):
Review OK.
I've tested this and everything seems to be fine.

PR #505, opened 2021-05-24
Implement server feature checking

Maxmystere:
(_I mostly did this feature for myself but i'm still sharing it if it's relevant for the project_)

Added a test when `Test Connection` button is pressed that will hide item in the menu drawer if the feature is not implemented on Server

Screenshot_20210524-232912


ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
```diff
@@ -359,6 +359,20 @@ public static synchronized String formatPercentage(int percent)
 		return Math.min(Math.max(percent,0),100) + " %";
 	}
 
+	/**
+	 * Return a Mark depending on boolean (❌ or ✓)
+	 *
+	 * @param value Is true or false
+	 * @return The corresponding mark.
+	 */
+	public static synchronized String boolToMark(boolean value)
```
tzugen: Doesn't need to be synchronized...

ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
```diff
@@ -606,6 +620,18 @@ public void onClick(DialogInterface dialog, int i)
 		}).show();
 	}
 
+	public static void showDialog(Context context, int icon, int titleId, String message)
+	{
```
tzugen, line 600: Try to use applicationContext() instead of passing around a context

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/EditServerFragment.kt
```diff
@@ -330,17 +338,65 @@ class EditServerFragment : Fragment(), OnBackPressedHandler {
                 pingResponse = subsonicApiClient.api.ping().execute()
                 ApiCallResponseChecker.checkResponseSuccessful(pingResponse)
 
+                updateProgress(String.format(progressString, "⌛", "⌛", "⌛", "⌛"))
```
tzugen: There is quite a bit of code duplication here. Maybe put the handling of which string to display into the UpdateProgress() function. And does this class ha e a viewmodel? If yes the actual connectio. Checking should go there...

tzugen (commented):
Hi I put some annotations in the code. I couldn't test it yet. Is the screen shot showing what you did?

tzugen:
Run ./gradlew -Pqc ktlintFormat
on your machine to fix most formatting problems

Maxmystere:
Yes my screenshot shows that i'm missing Chat and Shares because my server doesn't handle those requests

ultrasonic/src/main/java/org/moire/ultrasonic/util/Util.java
Maxmystere: I might need help on how to do that here

tzugen: I was actually wrong! Unlike toast, AlertDialog actually requires the Activity context!

But I cleaned up the methods a bit (see my commit, hope that was ok)

Maxmystere:
Thanks a lot for your help ! 
I also started thinking of a cleaner way to do testConnection that would even allow for a future async capability
I will do it soon if I find the time

tzugen:
I fixed the Ktlint issues, but the static analysis is showing some warnings. It's basically the same that I pointed at here:
https://github.com/ultrasonic/ultrasonic/pull/505#discussion_r638524883

Let me know if you need help fixing it.

tzugen:
I think it would be best to move the check function into serverSettingsModel, and then on the UI side (EditServerFragment) just query the model (`serverSettingsModel.hasChatFeature` or something like this).

You can have a look at how to run the query async with Kotlin coroutines, look at GenericListModel, lines 76-80. The loadFromServer method then uses `withContext(Dispatchers.IO) {` to run the query in a coroutine and don't block the main thread

nitehu:
This PR is a great idea! 
Actually Ultrasonic has a similar feature detecting the server version to find out the type of authentication to use, you can read about it in #253, and that also stores the detected version in the DB.

There was one problem we faced implementing that, namely that the users can upgrade (and rarely, but downgrade too) their servers and they expect Ultrasonic to follow that without adding the server again. 
So please just think it over before disabling a feature permanently for a server what should the user do to re-test and re-enable it. Maybe it is enough to add a dialog with something like "Ultrasonic found that your server doesn't support these features: [list]. These will be disabled for now, please run the tests again when the server is updated"

Maxmystere:
Actually I'm using the "Test Connection" button to refresh the server features so if you go and edit the server it will check again when you press "test connection" (maybe I should rename it)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/EditServerFragment.kt
Maxmystere: Updated
It's a bit cleaner now

nitehu:
@Maxmystere I've changed a few lines, please check if OK with you.
Also please let us know if you're planning further changes (for example you can move the PR to Draft while you aren't ready)

tzugen:
@nitehu @Maxmystere  before we merge, I would like to move some of the code into the model if that's ok for you, max...

Maxmystere:
Thanks a lot for that `isServerFunctionAvailable` function ! I tried a lot doing it but couldn't manage to fix some errors

I'm not an expert android developer so if you find anything you wanna fix before merging don't hesitate to do it 
And if I can be of any help tell me

tzugen:
I thought to try out this Kotlin Flow thing, to check for features concurrently, but I couldn't wrap my head around it, and I won't have much time this week, so I am fine with merging it as it is :+1:

tzugen (approved):

nitehu (approved):

PR #506, opened 2021-05-27
Kotlin services

tzugen:
This converts all remaining Music servic classes to Kotlin. This was a bit tricky, because the translation tool had problems inferring the nullability correctly, with a Java interface and Java and Kotlin implementations.

I also fixed a number of other lint, ktlint, and detekt issues and updated their baselines


tzugen:
The OfflineMusicService is more or less straight converted from Java. I didn't attempt to simplify the nested functions, because I want to rework the class later on with the database stuff anyway.

tzugen:
@nitehu I have tested all the functions that my server supports, but it would be good if you could test as well :)

nitehu:
This happens when I select an album from the Search screen:

```
D/NavigationActivity$onCreate: Navigated to org.moire.ultrasonic.debug:id/trackCollectionFragment
W/CommunicationErrorHandler$Companion: java.lang.NullPointerException
        at org.moire.ultrasonic.fragment.TrackCollectionFragment$updateDisplay$3.invokeSuspend(TrackCollectionFragment.kt:289)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7656)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
D/OpenGLRenderer: endAllActiveAnimators on 0xb27c27f0 (RippleDrawable) with handle 0xbd34ca10

```
The parentId is null here.
This also happens sometimes when I just browse the albums, and both in TrackCollectionFragment line 284 and line 289 may throw similar exceptions.

All other functions seem to work fine, but I'll continue testing on my phone.

tzugen:
Good catch :+1: 
It's fixed now.

nitehu (approved):
Seems to be working fine now.

PR #512, opened 2021-05-29
[DO NOT MERGE] Initial Test of Android Auto

SaintDubious:
This is not a real Pull Request, this is just a place for me to show some recent work, document the diff, and request suggestions for improvements.

This code represents an extremely basic start for getting Ultrasonic to work in Android Auto. It puts the ultrasonic icon in Auto, lets the user select from 10 albums, and then will play a song. This is not meant for actual release.


ultrasonic/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,8 @@
           package="org.moire.ultrasonic"
           android:installLocation="auto">
 
+    <uses-sdk android:minSdkVersion="20" android:targetSdkVersion="29" />
```
SaintDubious: I'm not actually sure what a good version is. This change, and all the rest in this file come out of [this doc here](https://developer.android.com/training/cars/media/auto) and [this one](https://developer.android.com/training/cars/media).

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/LocalMediaPlayer.kt
```diff
@@ -125,6 +125,10 @@ class LocalMediaPlayer(
     }
 
     fun release() {
+        // Calling reset() will result in changing this player's state. If we allow
+        // the onPlayerStateChanged callback, then the state change will cause this
+        // to resurrect the media session which has just been destroyed.
+        onPlayerStateChanged = null
```
SaintDubious, line 131: See the comment, I think this here is a fix for a bug that exists regardless of Android Auto. 

When the Service is being shutdown it calls `MediaPlayerService::onDestroy()` which calls this function here. When this function `reset()` below, it sets the play's state to IDLE. But the act of setting the state will call a callback, which will then start [re-creating the Media Session](https://github.com/ultrasonic/ultrasonic/blob/develop/ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt#L371). Then the `MediaPlayerService::onDestroy()` will resume and continue destroying the Media Session.

This is all bad behavior for the existing app, but for Android Auto we have to set the session token, which can only be done once. All this destroying and re-creating ends up attempting to set it twice, which results in an exception. By setting `onPlayerStateChanged` to `null` here we remove the callback and allow this thing to die gracefully.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -49,8 +57,7 @@ import timber.log.Timber
  * while the rest of the Ultrasonic App is in the background.
  */
 @Suppress("LargeClass")
-class MediaPlayerService : Service() {
-    private val binder: IBinder = SimpleServiceBinder(this)
+class MediaPlayerService : MediaBrowserServiceCompat() {
```
SaintDubious, line 54: Making this class inherit from `MediaBrowserServiceCompat` is what's suggested in [this doc here](https://developer.android.com/guide/topics/media-apps/audio-app/building-a-mediabrowserservice).

The binder is no longer needed as the `MEdiaBrowserServiceCompat::onBind()` callback is now correct.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -62,20 +69,25 @@ class MediaPlayerService : Service() {
     private val mediaPlayerLifecycleSupport by inject<MediaPlayerLifecycleSupport>()
 
     private var mediaSession: MediaSessionCompat? = null
-    private var mediaSessionToken: MediaSessionCompat.Token? = null
     private var isInForeground = false
     private var notificationBuilder: NotificationCompat.Builder? = null
 
+    val executorService: ExecutorService = Executors.newFixedThreadPool(4)
```
SaintDubious: I use a thread pool to get network calls off the main thread. See below

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -62,20 +69,25 @@ class MediaPlayerService : Service() {
...
+    private val MEDIA_BROWSER_ROOT_ID = "_Ultrasonice_mb_root_"
+    private val MEDIA_BROWSER_ALBUM_LIST_ROOT = "_Ultrasonic_mb_album_list_root_"
+    private val MEDIA_BROWSER_ALBUM_PREFIX = "_Ultrasonic_mb_album_prefix_"
+    private val MEDIA_BROWSER_EXTRA_ENTRY_BYTES = "_Ultrasonic_mb_extra_entry_bytes_"
+
     private val repeatMode: RepeatMode
         get() = Util.getRepeatMode()
 
-    override fun onBind(intent: Intent): IBinder {
-        return binder
```
SaintDubious, line 73: This implementation actually doesn't work and caused me a lot of pain. The default provided by `MediaBrowserServiceCompat` works correctly. No need to override it

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -62,20 +69,25 @@ class MediaPlayerService : Service() {
...
-    }
-
     override fun onCreate() {
         super.onCreate()
 
+        updateMediaSession(null, PlayerState.IDLE)
+        mediaSession!!.isActive = true
```
SaintDubious: Previously this service delayed creating of the Media Session for as long as possible. For Android Auto we need to create it as soon as possible, inside the `onCreate`. The doc [here](https://developer.android.com/training/cars/media#browser_workflow) explains this. Thankfully the existing function works well, except IDLE sets the `mediaSession.isActive` to false.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -136,6 +148,99 @@ class MediaPlayerService : Service() {
         }
     }
 
+    override fun onGetRoot(
```
SaintDubious, line 139: This is the key part of the browsing interface. Usage is spelled out [in this doc](https://developer.android.com/training/cars/media#onGetRoot). Note that the root itself isn't actually displayed in Android Auto.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -136,6 +148,99 @@ class MediaPlayerService : Service() {
...
+    }
+
+    override fun onLoadChildren(
+        parentMediaId: String,
+        result: MediaBrowserServiceCompat.Result<List<MediaBrowserCompat.MediaItem>>
+    ) {
+
+        val mediaItems: MutableList<MediaBrowserCompat.MediaItem> = mutableListOf()
+
+        if (MEDIA_BROWSER_ROOT_ID == parentMediaId) {
```
SaintDubious: This part would be like the Main Menu. As you can see I only made one entry called "Browse Albums." According to the Android Auto docs there should only be like 4 entries in the root, so we'd have to choose wisely. There's also newer UI that has some tabs, which we should opt into and start to use. But what should be on these tabs and main menu? This part feels like there needs to be some community agreement.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -136,6 +148,99 @@ class MediaPlayerService : Service() {
...
+                    MediaBrowserCompat.MediaItem.FLAG_BROWSABLE
+                )
+            )
+        } else if (MEDIA_BROWSER_ALBUM_LIST_ROOT == parentMediaId) {
+            executorService.execute {
+                val musicService = getMusicService()
+
+                val musicDirectory: MusicDirectory = musicService.getAlbumList2(
+                    "alphabeticalByName", 10, 0, null
+                )
```
SaintDubious: This part happens when someone clicks on "Browse Albums" from the main menu. As you can see here I just download 10 albums using one call I know works from the `musicService`. Usually there are checks for ID3 settings and other things. This should be more robust. 

Also, do we even want to download? Maybe Android Auto should only use Offline Mode? Pretty much guaranteed users won't be using their home WiFi here. Also, in a car we don't really want to show hundreds of albums in a user's collection.

This is another case where we need some community agreement

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -136,6 +148,99 @@ class MediaPlayerService : Service() {
...
+                    id, "", false
+                )
+                for (item in albumDirectory.getAllChild()) {
+                    var extras = Bundle()
+
+                    var baos = ByteArrayOutputStream()
+                    var oos = ObjectOutputStream(baos)
+                    oos.writeObject(item)
+                    oos.close()
+                    extras.putByteArray(MEDIA_BROWSER_EXTRA_ENTRY_BYTES, baos.toByteArray())
```
SaintDubious: This is where we add the song list to display after someone selects an album. Note that `item` implements `Serializable` so I should just be able to call `extras.putSerializable(...)` but when I tried that it crashes. Not sure if we don't implement something in `Serializable` correctly. I cheat and use a byte array, I'd need to clean this up.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -816,6 +921,32 @@ class MediaPlayerService : Service() {
                 Timber.v("Media Session Callback: onPlay")
             }
 
+            override fun onPlayFromMediaId(mediaId: String?, extras: Bundle?) {
+                super.onPlayFromMediaId(mediaId, extras)
+
```
SaintDubious, line 835: This here is the last piece of magic. Once a user has selected a song from the list, this function will be called. The `extras` object contains all the data I need to set up the download and play.

Took me forever to find this bit, turns out it's documented right [here](https://developer.android.com/training/cars/media#playback-commands). A lesson to read all the way to the bottom of even long documents :)

SaintDubious (commented):
Here are the things I wanted to point out about this code. Suggestions more then welcome, in fact I'm hoping some folks have some good ideas about which direction to go.

SaintDubious:
One final thought and then I'll stop typing for the night. While the Android Docs are the ones who tell us to implement the `BediaBrowserServiceCompat` in this way, I don't really love it. We're adding a whole bunch of new code (all the browsing and menu items) that are only used by Android Auto. It's gonna make the normal app just that much less clear to read. I'd like to find a way to separate the Android Auto stuff, but not super clear on how to do it. I **could** register a separate service specifically for Android, but then I'd have to re-create all the Media Session stuff, which the existing service already does.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
tzugen: And for Auto you need it to be active?
If I remember correctly, if it is active it will show lockscreen controls, which we wanted to avoid as lo g as we don't actively play something

tzugen:
Hi! Thanks for this work!

I think it is fine to change the implementation of the Service as you have done, regarding from which to inherit, and the binder stuff.

But I share your concerns from your last post. If we start putting UI code now I to the service it will get quite messy. Maybe this can at least go into a util class, and then we only call one or a few methods from the Util class to setup the menus? 

Also regarding the threads, I think you should be able to reuse the respective ViewModels to gather the data. Like ArtistListModel etc. Then you don't have to reinvent the wheel there. The model is also using Kotlin Coroutine, which are a more lightweight alternatives to spanning threads.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
nitehu: I think we must support online mode. I always use mobile data to play music in my car (that's why I have a subsonic server in the first place, so I can access my collection on the go). I think I'm not alone with this, also unlimited data plans are becoming more common...
I agree showing many-many albums at a time is pointless, I can't imagine browsing for minutes while driving. But the search (also voice search) becomes more important...

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
SaintDubious: Looks like it's not needed. This one must have been an artifact of my previous flailing, which I did a lot of.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
SaintDubious: Okay cool, this is valuable feedback. I was burned by the endless download bug, so I tend to download all my music while at home and then only ever run Ultrasonic in "Offline Mode" when I leave the house. I wasn't sure what was common. Okay, I'll pursue ways to make it usable assuming hundreds of albums accessible online.

tzugen:
Regarding Separation of concerns:

Maybe you could create a Service "CarService" that extends MediaBrowserService and then have the main service extend that. Then all the code regarding car stuff can stay in one class. Or if that is not possible the other way around, CarService extending the main service.

SaintDubious:
Okay, moving forward a bit here. I refactored the Auto specific stuff into a separate class. I'm calling this technique "Hard Coded Dependency Injection," but in theory there could be an interface and all the usual niceties. For now it's just a util class to dump all the car-specific stuff into. Makes me feel better about the organization

@tzugen I like the idea of re-using the Models, but they seem to depend on the `SwipeRefreshLayout` which I understand to be a UI element. The service has no UI component as Android Auto itself handles all the UI bits. In fact even Kotlin co-routines I understand to depend on having more UI infrastructure then just vanilla threads (I think I read something about that somewhere). This is now the sticking point for me. I am basically cutting and pasting and re-inventing the whole subsonic client side, which isn't a good thing to do. But I don't see any existing code that fits nicely. Any guidance would be appreciated. Maybe I just need to refactor the models? Or maybe it's okay to have a `SwipeRefreshLayout` with no UI?

tzugen:
It is always when you try to reuse code, that you find the places where UI and logic have not been separated well! 

We should not update the UI from the model anyway. It is just used to display the spinning loading animation.

The View should observe the livedata and enable/disable the animation based on the loading state.

Good catch :) 

07.06.2021 00:36:17 James Wells ***@***.***>:

nitehu:
> Or maybe it's okay to have a SwipeRefreshLayout with no UI?

Oh the SwipeRefresh is just an old remnant of a progress indicator that we haven't refactored yet, as tzugen mentioned :) I'd suggest that temporarily it can be made nullable in the models easily, so you can reuse the models and progress with more important parts of your code...

> The View should observe the livedata and enable/disable the animation based on the loading state.

That would be absolutely the best solution, but currently the error handling also uses the context of the SwipeRefresh :( so it seems to separate the UI, the error handling must be lifted off too...

tzugen:
Yes, making it nullable and maybe adding and issue here on GH to properly remove it later would be the easiest :) 

07.06.2021 08:24:20 Nite ***@***.***>:

SaintDubious:
Thanks @tzugen and @nitehu ! That's great info and validates what I thought I was seeing. It also gives me a direction to try out. I'll see how making that field nullable works out.

SaintDubious:
I'm kind of shocked to report that I got a working version. It's pretty darned minimal, but it works. Sadly, not all of the models could be re-used (so I had to retain the `ExecutorService` thing for some stuff). And even when I did re-use some stuff, the mixture of "Observing" the `LiveData` for the lifetime of the class does not really fit elegantly with only needing to fill the `results` of one callback (see my `ArtistListObserver` and `AlbumListObserver` classes). Also the `LiveData` thing seems to want a View aware observer, so I wasn't able to just create a local observer (which would have fit more elegantly).

So yeah, it still feels a bit awkward, but I'm hoping it can server as a starting point for a minimal release.

Trying to set up a test environment for this thing is a mission in itself. Does any reviewer have a way to test this? I can provide details on how I got it working (basically I did what's described here: https://developer.android.com/training/cars/testing ).

The actual integration just provides 3 top level views: Recent, Albums, and Artists. You can't play a whole Album as a single unit, but when you select a song from an Album it will download and queue up that song and all of the rest of the songs on the Album. I've never used Android Auto before, so I don't really know the best way to present this. The actual screen for playing songs is nice, it has album artwork and the expected play controls (I can't take credit for that, it "just worked" somehow).

nitehu:
I"ll try to check this out next week. I don"t yet have an emulator for it (neither a real device), good to know its a pain to set one up :)

SaintDubious:
Ugh, I'm sorry to say that I have still never gotten a dev environment to work without a real phone in the mix. I've tried a number of different hardware and OS versions on emulators and none of them can get a version of Android Auto to speak with the Desktop Head Unit (version 1.x or 2.x). If you can figure something out please let me know. Having to develop with a physical phone is a pain.

ultrasonic/src/main/AndroidManifest.xml
nitehu: Actually this seems to be unnecessary, see [this](https://stackoverflow.com/questions/19997509/android-studio-why-are-minsdkversion-and-targetsdkversion-specified-both-in-and)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
nitehu: Actually I think it would be better to create a separate MediaBrowserService in Ultrasonic, which would then call the MediaPlayerService for playback related things, but would contain the browsing related logic.
(Also, I suggest to take a look on the DSub code just for hints/ideas, that is another subsonic fork so not too different from Ultrasonic)

nitehu (commented):
So I've tried out this implementation, very nice work!
Unfortunately I too haven't found a way to run it without a physical phone...

At least I could compare it to other players, and found the following:

Screenshot 2021-06-22 160431

It seems it is common to add a menu-like structure to the root level instead of the media items. On the previous shot, the horizontal items are very similar to Ultrasonic's menu, and the items in the "Library" are similar to Ultrasonic's Main screen. Something like this would be ideal IMHO.

I've also tested what can Android Auto do when there are many-many albums on the server, and it froze :( Some kind of paging should be implemented ideally, but I'm not sure how...

Also, the player screen looks like this:

Screenshot 2021-06-22 160510

Here the leftmost icon displays the playlist, and the three dots on the right brings up a menu with shuffle/repeat options.
I haven't the slightest idea how they did that :D

I don't think all this must be implemented for a minimal release... Browsing IMHO is not so important than solid playback controls.
Also it would be nice to separate out the media browsing and the playback in the code...

SaintDubious:
Thanks @nitehu as always for the review. Been a bit busy these days, but schedule should lighten up soon. I'll see how it looks to rearrange the inheritance. Thanks also for the tips on the menu, I see that we do have 3 major categories: Music, Songs, Albums on the main menu, so that's a good idea to copy that layout. Anyway, will update soon.

SaintDubious:
I have some free time to pick this up again. The last push was just a merge. I'm gonna try reversing the inheritance in a new branch so we can see if we like it.

nitehu:
Android Auto will be implemented with a different approach in #537

PR #514, opened 2021-06-02
Fix bugs in new image loader and make it default.

### LINKED ISSUE #368: Empty id in getCoverArt
## sharing of playlists doesn't work

Trying to share an existing playlist shows an error "requested data not found"

### Steps to reproduce

See above.

## System information

 * **Ultrasonic version**: 2.17.2
 * **Android version**: 8.0.0
 * **Device info**: HUAWEI MediaPad M5 lite

### LINKED ISSUE #380: App crashes when scrolling through albums
## Problem description

The app crashes when I scroll through the albums and go to the ~2-3 first pages. Note that I use the new image loader and I display 25 albums per page.

### Steps to reproduce

1. Open the app
2. Open the list of albums
3. Scroll down to get the next page button
4. Click on next page button to go to the next page
5. Repeat steps 3-4 a few times (3 times is enough)
6. App crashes

## System information

 * **Ultrasonic version**: *2.17.2*
 * **Android version**: *Android 8.1.0*
 * **Device info**: *Xiami, Mi A2*

## Additional notes

Here is the [ultrasonic-stacktrace.txt] file that's being dumped on the phone when it crashes.

[ultrasonic-stacktrace.txt]: https://paste.sr.ht/~tomleb/480b5fc4252856ff0b952b3a0f683061a7de746e

I have tested without the new image loader and it does not crash.

Unrelated, but when can we see 2.17.3 released on F-Droid?

### LINKED ISSUE #511: Make Service.getCoverArt() require a non-null Entry
After #506 has been merged:

Make Service.getCoverArt() require a non-null Entry
Should fix #368

tzugen:
There were two main bugs in the new image loader:
1. Id was checked for nullability, but it is actually an empty string in most cases where there is no Cover art.
This lead to queries without id set (fixes #380)
2. Size was not respected by the new image loader. This lead to bad performance.

The caching strategy for Offline use is still the same. Later on we could try to implement on disk caching using Picasso,
but since our solution works fine, it's not urgent


tzugen:
Closes #511

tzugen:
I noticed it doesnt work in offline mode. 
I woud need to use `FileUtil.getAlbumArtBitmapFromDisk` in CoverArtRequestHandler,
but I can't since this would create a circular dependency between subprojects.

@nitehu Do you have an idea how to solve this?

tzugen:
And it actually makes more sense to stay with our cache implementation, because with OkHttp caching we would cache the most recently accessed cover arts, where now we cache specifically those which are available offline.

nitehu:
> I woud need to use FileUtil.getAlbumArtBitmapFromDisk in CoverArtRequestHandler,
but I can't since this would create a circular dependency between subprojects.

So there are two ways to fix this:
- I'm not sure why the subsonic-api-image-loader is in a separate module... it can be moved to the ultrasonic module without any hassle
- But maybe modularizing the app is the way forward, in this case it may be solved by creating a "common" or similar module in which we can include things we use from multiple other modules. I can imagine (parts of (?)) Utils, FIleUtils and other small, reusable helper classes can be moved there. 

I'd vote for the latter approach, but that seems to be more work.
What do you think?

Edit: Actually there is a third way... As I see the modules inside `core` doesn't necessarily know that they are running inside an Android application. To keep it this way, an "offlineLoader" interface can be created inside the image loader, for which we can inject an implementation with DI from the ultrasonic module.

tzugen:
Ah ok, I saw your last comment too late. I moved it inside the main component, because it's not just abstractly loading something, but it is loading images directly into ImageViews. 
@nitehu it's ready for review now.

nitehu:
I've found a totally unrelated bug, which thrown an exception on me during testing... Please check if the fix is OK with you.
I'll continue the tests.

nitehu:
There was another one: after changing from a song with existing cover art to one which has none set, the old cover art remained visible. The old image loader solved this by setting the unknown image for the view directly in these cases, so I've copied a similar solution.

nitehu:
Ok there is a last (I hope) problem, the cover arts shown on the player screen are sometimes very pixelated for me.
It seems sometimes the default sized images were retrieved from the cache instead of the large ones.

tzugen:
Thanks for fixing the issues directly :)
I extended your last commit, because on disk we are caching only the full sized image (see the commit message for more info)

tzugen:
Now the function getCoverArt in the MusicService was used only by the Downloader anymore and getAvatar was not used at all anymore.

So I cleaned up these functions, and moved the parts which are responsible for downloading the cover art when the user wants to download an Album for offline use into the ImageLoader class.

nitehu (approved):
Review OK, tests OK
This can be merged if you don't plan other modifications

PR #517, opened 2021-06-09
Cleaner separation of API result handling.

tzugen:
* CallResponseChecker functionality are now Kotlin Extensions
* Removed unfitting shortcuts from SubsonicAPIClient
* Increase timeout to get a video url
* Fix a bug, that the Rest API version was checked twice on each call


tzugen:
@nitehu 
The CoverArtRequestHandlerTest is failing, because it seems it doesn't like the mocked function anymore since I removed the shortcut. I can't figure out how to fix it.Could you have a look? It would be a pity to have to readd these functions, just to get the test to pass

tzugen:
(This was all issues that I saw when working on the last PR)

nitehu:
Oh I see you've found it... :)
I also found that Mockito is not so forgiving when mocking multiple calls like that. Unfortunately I haven't got a better solution than yours - unless of course someone wants to completely rewrite the tests...

nitehu (approved):
Review OK.
I've tested the functions for a few minutes and everything seems to work as intended.

PR #519, opened 2021-06-10
Better experience for new users

tzugen:
With this PR our database is prepopulated with a demo server running ampache (see https://github.com/ampache/ampache/issues/2938 )

This makes it more welcoming for new users because they can check out the app before getting a server themselves.
Also sets the default preferences for new users to use ID3 display. This is what most people are used to from iTunes or Spotify.

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/activity/NavigationActivity.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/activity/NavigationActivity.kt
@@ -326,7 +326,7 @@ class NavigationActivity : AppCompatActivity() {
                 .setTitle(R.string.main_welcome_title)
                 .setMessage(R.string.main_welcome_text_demo)
                 .setNegativeButton(R.string.main_welcome_cancel) { dialog, _ ->
-                    // Got to the settings screen
+                    // Go to the settings screen
                     dialog.dismiss()
                     findNavController(R.id.nav_host_fragment).navigate(R.id.settingsFragment)
                 }
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/activity/NavigationActivity.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/activity/NavigationActivity.kt
@@ -29,7 +29,6 @@ import androidx.preference.PreferenceManager
 import com.google.android.material.navigation.NavigationView
 import org.koin.android.ext.android.inject
 import org.koin.androidx.viewmodel.ext.android.viewModel
-import org.koin.java.KoinJavaComponent.inject
 import org.moire.ultrasonic.R
 import org.moire.ultrasonic.data.ActiveServerProvider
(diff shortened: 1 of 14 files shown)


tzugen:
@nitehu @ogarcia 
What do you think of this?
Maybe it's worth discussing because it changes the behaviour of the app..

nitehu:
> our database is prepopulated with a demo server running ampache

Speaking for myself, I very much like this idea! But trying to see through a UX viewpoint (which I'm probably bad at):
- con: I'm not sure how experienced the users are, but I'm afraid someone may think that they can use that server for anything else than just a demo (I'd hate to see "ok but how can I add more songs" comments popping up 😂). I'd write "this is just a demo, please remove it and use your own server" with giant letters everywhere...
- pro: Yes, users can try Ultrasonic out - but I think most of the users will search for an app for their existing server. However, maybe if they can see how well Ultrasonic works with a _good_ server, Ultrasonic can spare itself some negative reviews...

> sets the default preferences for new users to use ID3 display

I agree this is a good idea, maybe power users prefer to browse by path but they will find the setting and switch it back.

tzugen:
Ah, yes that could happen as well 😂

I forgot to write, another motivation was also that I wanted to try using Firebase Robotests ( https://firebase.google.com/docs/test-lab/android/robo-ux-test
) to automatically find exceptions, and there it would be great if the app was pre-configured to have a server.

11.06.2021 11:29:23 Nite ***@***.***>:

>> our database is prepopulated with a demo server running ampache
>> 
> Speaking for myself, I very much like this idea! But trying to see through a UX viewpoint (which I'm probably bad at):
> 
* > con: I'm not sure how experienced the users are, but I'm afraid someone may think that they can use that server for anything else than just a demo (I'd hate to see "ok but how can I add more songs" comments popping up 😂 ). I'd write "this is just a demo, please remove it and use your own server" with giant letters everywhere...
* > pro: Yes, users can try Ultrasonic out - but I think most of the users will search for an app for their existing server. However, maybe if they can see how well Ultrasonic works with a /good/ server, Ultrasonic can spare itself some negative reviews...

tzugen:
(I used Robotests a lot to polish another app I am maintaining ( BodhiTimer))

ogarcia:
Maybe instead of have a demoserver preconfigured is better an option to add the demoserver in server administration. This is better because you can add this demoserver at any time and can be used to test connection problems (if demoserver works but your server don't then the problem is in your server)

tzugen:
Ok, thanks for the feedback. I now made it optional to add the demo server in the welcome screen, and made the wording clear that users will need their own server to really use the app. 
I didn't add it in the server settings screen, because there was no menu that i could easily insert it, and I didn't want to add extra UI for this.

nitehu (approved):
I think this is fine now.
- It is reasonably clear that the added server is just a demo
- Adding it is optional, so just a touch of a button and Ultrasonic works exactly as before

(there were 2 typos which I fixed, I hope it is okay with you)

tzugen:
Yes, i like the collaborative work on PRs :)

PR #520, opened 2021-06-12
Remove flash

### LINKED ISSUE #516: Remove flash stuff from video
In VideoPlayerType.java we are advertising a non-free Video Player, seemingly just to make it possible to playback flash videos.
We are in the year 2021. Flash is dead.

tzugen:
(It was broken anyway). Closes #516 
Also removes advertisment of non-free video player.


nitehu (approved):
Review was OK!
I've never used the video settings in Ultrasonic... I haven't seen flash videos in years, also MX player just works with the default setting - as any android player should.

PR #525, opened 2021-06-14
Properly generate the Video stream url, without actually making a request

### LINKED ISSUE #524: Video playback fails for large videos
## Problem description

Video playback for large videos fails now with an OutOfMemoryException, as if Ultrasonic would try to load the whole video into memory. 
I think this should work by only retrieving the video stream url from the server and passing it to a video player app, but somewhere there is a bug in this process.

### Steps to reproduce

- Go to Videos
- Click a large file (~over 20 minutes)
- Ultrasonic crashes

## System information

 * **Ultrasonic version**: 2.21.0
 * **Android version**: Android 10
 * **Device info**: Samsung Note 9
 * **Server**: Airsonic 10.6.2

## Additional notes

Logcat: 
[exception2.txt](https://github.com/ultrasonic/ultrasonic/files/6649085/exception2.txt)

tzugen:
See comments in commit..
Fixes #524


nitehu (approved):
Works perfectly, looks a bit strange... but I'm almost certain there isn't an easier and nicer way to fix this :)
Also, video playback is a lot snappier now.

PR #526, opened 2021-06-16
Fix no server call when starring song from notification

Maxmystere:
I mostly only use the notification when using the app and I noticed recently the the star button didn't work

Tell me if there is any improvement to do with that PR


tzugen (approved):
It's ok. We always need to somehow wrap the call when making a server call.
And MediaPlayerController is not yet converted to Coroutines, so it's totally fine to make a thread.

Thanks!

PR #529, opened 2021-06-19
Fix save playlist

### LINKED ISSUE #528: Saving playlists does not work
## Problem description

Trying to save a playlist results in a toast with an error message about a non null parameter being null. The playlist is not saved. 

### Steps to reproduce

Queue some songs and try to save the queue as a playlist.

## System information

 * **Ultrasonic version**: 2.21.0
 * **Android version**: 11
 * **Device info**: Google Pixel 4a

tzugen:
This fixes #528 as well a a long standing leak (LocalMediaPlayer was holding a reference to the Seekbar, even when the view was gone).


nitehu (approved):
Tested this, saving playlists now works correctly.
I've also tried to press everything else to find bugs in the kotlin conversion, but it seems to work fine.

PR #530, opened 2021-06-19
Use DiffUtil for better performance when refreshing the data..

tzugen:
See https://bignerdranch.com/blog/efficient-lists-with-diffutil-and-listadapter/ for more info on this


tzugen:
This is the nice thing with these bases classes. When I improve the GenericRowAdapter all the changes are automatically in Artist and album List etc.

tzugen:
I will add another commit today

tzugen:
Now when you select an Album and then naviagte back, you end up at the scroll position where you left of, and not on the top.
This was annoying for long lists!

nitehu (approved):
Everything seems to be OK.

PR #537, opened 2021-07-04
Added Android Auto support

### LINKED ISSUE #229: Cannot use Ultrasonic with Android Auto
## Problem description

Ultrasonic is not supported by Android Auto (or the other way 'round)

### Steps to reproduce

Try to use Ultrasonic with Android Auto

## System information

Ultrasonic version: 2.5.0
Android  * **7.1.1**: *32.4.A.1.54*
Sony Z3+

SaintDubious:
This is a different approach to https://github.com/ultrasonic/ultrasonic/pull/512

In this version instead of turning the existing MediaPlayerService into something that can work for Auto (mainly by bolting on the browsing capability) we are creating a new Service, AutoMediaPlayerService, which uses the MediaPlayerService to handle the MediaSession.

I'm not able to make this work and I have doubts about it. The existing MediaPlayerService makes the assumption that it is a proper Service, and using it instead as a member of a Service feels dangerous. The immediate source of the crash is in the call to `createNotificationChannel(...)` when it attempts to call `getSystemService(NOTIFICATION_SERVICE)` I get a null object reference. But even getting around that my big concern is that this code needs to add an override of `onPlayFromMediaId(...)` in the private `mediaSession` inside the `MediaPlayerService` which would be a big breach of encapsulation. Ideally the MediaSession and the LocalMediaPlayer could be refactored into some other class which both MediaPlayerService and AutoMediaPlayerService could use, but trying to untangle that is daunting.

Having said all that, I do prefer to have a separate service for Auto as it means we can change the two services independently as I assume they'll have different objectives going forward. So maybe that refactor is worth the attempt.

@nitehu what do you think?


nitehu:
Yep it seems you're absolutely right. I'll experiment with this when I have more time, until then it seems you're the one with the most experience with Android Auto :)

What do you think, is it necessary for the playback and the browsing to be in the same service? I think it is enough if they are entirely separate - MediaBrowserService just returns the playable items, and MediaPlayerService handles the playback.
Maybe the MediaBrowserService can start a real MediaPlayerService (e.g. by calling MediaPlayerService.getInstance() in its onCreate)?

The onPlayFromMediaId(...) is a tough nut... as you've said it is in an entirely wrong place deep inside Ultrasonic where the MediaId has no meaning at all. Ideally it would add the selected song / album to the playlist and would start playing it, but that functionality is in a layer above the MediaPlayerService.
My only guess is currently that we should create a new intent from it and handle it elsewhere.

What do you think?

SaintDubious:
I think I'm gonna see just how much code needs to be copied and pasted. I don't understand the bulk of what's going on in the current Service, maybe it's not needed for Auto? Maybe I can do a clean break without requiring too much duplication. </Extreme_Optimism>.

SaintDubious:
I'm afraid I've come up short. I have 2 distinct failures under my belt now.

I gave the `MediaPlayerService::getInstance()` trick a try. The problem for this comes down to order of service startup. `getInstance` works by creating the service and then sleeping a bit to give it time for its `onCreate()` to fire (and set the static `instance` variable). If I try this while my new service is being created (either in the constructor or my new `onCreate()`) the system won't schedule `MediaPlayerService::onCreate()` until **after** my own is done. So the end result is that `getInstance()` does **construct** a new instance of `MediaPlayerService`, but it doesn't finish creating it until too late. I need for it to finish creation (and critically set the session token) before my own `onCreate()` is done. There may be a way to work around this, but I generally still worry that implementing a Service by depending on a different Service is gonna cause issues.

I also attempted to just pull out the parts that we need into a separate Utility class, but that just got too difficult. I know I'd need the mediaSession, downloader, and localMediaPlayer, but I have no idea if I'd need the rest. Trying to tease them all apart is hard when I don't know what most of them do. I think this is still probably the correct approach, but I'm afraid it's beyond my reach at the moment. Maybe when I'm feeling another wave of optimism I'll give it another try

nitehu:
It seems I've just made this work with 2 separate services. My code is very minimalistic for now, and I don't want to mess up your work... I'll try to merge up something usable from the two tomorrow.

ultrasonic/src/main/AndroidManifest.xml
```diff
@@ -51,6 +59,16 @@
             android:exported="false">
         </service>
 
+        <service
+            android:name=".service.AutoMediaBrowserService"
```
nitehu, line 65: I've added this new service, but left the AutoMediaPlayerService in its place. I'm not sure if that should be rewritten entirely or if we can copy and reuse some parts from it.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/AutoMediaBrowserService.kt
```diff
@@ -0,0 +1,186 @@
...
+            override fun onMediaSessionTokenCreated(token: MediaSessionCompat.Token) {
+                Timber.i("AutoMediaBrowserService onMediaSessionTokenCreated called")
+                if (sessionToken == null) {
+                    Timber.i("AutoMediaBrowserService onMediaSessionTokenCreated session token was null, set it to %s", token.toString())
+                    sessionToken = token
+                }
+            }
+
+            override fun onPlayFromMediaIdRequested(mediaId: String?, extras: Bundle?) {
+                // TODO implement
```
nitehu: This will be called from the MediaSession.
Currently I imagine this should add the selected item(s) to the playlist as next track(s) and start playing it. 
This is debatable, but if the user has already loaded a playlist, I'd just add the selected tracks to it without removing anything. For me the main use case in the car is that some music always plays in the background, like a radio - if I select some songs directly, I want to listen to those, and then get back to radio mode :)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/AutoMediaBrowserService.kt
```diff
@@ -0,0 +1,186 @@
...
+                Timber.i("AutoMediaBrowserService onPlayFromSearchRequested called")
+            }
+        }
+
+        mediaSessionEventDistributor.subscribe(mediaSessionEventListener)
+
+        val handler = Handler()
+        handler.postDelayed({
+            Timber.i("AutoMediaBrowserService starting lifecycleSupport and MediaPlayerService...")
+            // TODO it seems Android Auto handles autostart, but we must check that
```
nitehu: So this here has a big problem.
I'm not sure how can we make the AndroidAuto logic to work with how Ultrasonic works currently.

When Ultrasonic is started, it usually loads up a playlist, and resumes playback if it wasn't completely stopped before.
If it was stopped, the playback can be started from the Now Playing screen.

However, when Ultrasonic is stopped, the Now Playing is not visible at all in Android Auto! I think because it uses the MediaSession, which is without a current track, and inactive in this case. So in Android Auto the playback only can be started after selecting a new something to play.

To solve this nicely, I think we must rethink how this should work and maybe refactor the MediaSession handling.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/AutoMediaBrowserService.kt
```diff
@@ -0,0 +1,186 @@
...
+    ): BrowserRoot? {
+        Timber.i("AutoMediaBrowserService onGetRoot called")
+
+        // TODO: The number of horizontal items available on the Andoid Auto screen. Check and handle.
+        val maximumRootChildLimit = rootHints!!.getInt(
+            MediaConstants.BROWSER_ROOT_HINTS_KEY_ROOT_CHILDREN_LIMIT,
+            4
+        )
+
+        // TODO: The type of the horizontal items children on the Android Auto screen. Check and handle.
```
nitehu: These were in the sample codes, I'm not sure wether we want to handle them

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/AutoMediaBrowserService.kt
```diff
@@ -0,0 +1,186 @@
...
+        extras: Bundle?,
+        result: Result<MutableList<MediaBrowserCompat.MediaItem>>
+    ) {
+        super.onSearch(query, extras, result)
+    }
+
+    private fun getRootItems(result: Result<MutableList<MediaBrowserCompat.MediaItem>>) {
+        val mediaItems: MutableList<MediaBrowserCompat.MediaItem> = ArrayList()
+
+        // TODO implement this with proper texts, icons, etc
```
nitehu: This is just a test to see how the horizontal menu appears

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/AutoMediaBrowserService.kt
```diff
@@ -0,0 +1,186 @@
...
+            )
+        )
+
+        result.sendResult(mediaItems)
+    }
+
+    private fun getAlbumLists(result: Result<MutableList<MediaBrowserCompat.MediaItem>>) {
+        val mediaItems: MutableList<MediaBrowserCompat.MediaItem> = ArrayList()
+
+        val description = MediaDescriptionCompat.Builder()
```
nitehu: And yep, we should add the code here to load the real media items

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/LocalMediaPlayer.kt
```diff
@@ -167,7 +171,7 @@ class LocalMediaPlayer(
             val mainHandler = Handler(context.mainLooper)
 
             val myRunnable = Runnable {
-                onPlayerStateChanged!!(playerState, currentPlaying)
+                onPlayerStateChanged?.invoke(playerState, currentPlaying)
```
nitehu, line 179: This was an old bug. onPlayerStateChanged can be null here, because it may be gone in the time the execution arrives here.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -95,6 +96,12 @@ class MediaPlayerService : Service() {
 
         localMediaPlayer.onNextSongRequested = Runnable { setNextPlaying() }
 
+        // TODO maybe MediaSession must be in an independent class after all...
```
nitehu: I think one of the next steps could be to refactor the MediaSession to its own handler class if that's possible...

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -538,7 +552,8 @@ class MediaPlayerService : Service() {
             }
         }
 
-        playbackState.setState(state, PlaybackStateCompat.PLAYBACK_POSITION_UNKNOWN, 1.0f)
+        // TODO playerPosition should be updated more frequently (currently this function is called only when the playing track changes)
```
nitehu: This is necessary for the Android Auto to show playback position in the track during playback.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -549,6 +564,14 @@ class MediaPlayerService : Service() {
         // Set Active state
         mediaSession!!.isActive = isActive
 
+        // TODO Implement Now Playing queue handling properly
+        mediaSession!!.setQueueTitle("Now Playing")
```
nitehu: I just wanted to try out how the Now Playing queue is displayed and handled in Android Auto, so this code is just a test, but works nicely.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -799,6 +822,7 @@ class MediaPlayerService : Service() {
 
         mediaSession = MediaSessionCompat(applicationContext, "UltrasonicService")
         mediaSessionToken = mediaSession!!.sessionToken
+        mediaSessionEventDistributor.RaiseMediaSessionTokenCreatedEvent(mediaSessionToken!!)
```
nitehu: This solution is actually [suggested by Google](https://developer.android.com/reference/kotlin/android/support/v4/media/session/MediaSessionCompat)

> the owner of the session may pass its session token to other processes to allow them to create a MediaControllerCompat to interact with the session

That's why I think the MediaSession should be a thing of its own, which is used by the MediaBrowser and the MediaPlayer

nitehu (commented):
So this is where I'm at now...
It works (for me), but I still have some problems with the MediaSession. Ultrasonic handles this currently in a way that is hardly compatible with Android Auto. I'll try to solve the problems as the next step...

I've wanted to make the playback part with the Now Playing screen work first. Currently the media browsing is just a dummy implementation, but I think the code can be copied over from #512 later...

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/AutoMediaBrowserService.kt
```diff
@@ -0,0 +1,186 @@
...
+            override fun onPlayFromSearchRequested(query: String?, extras: Bundle?) {
+                // TODO implement
+                Timber.i("AutoMediaBrowserService onPlayFromSearchRequested called")
+            }
+        }
+
+        mediaSessionEventDistributor.subscribe(mediaSessionEventListener)
+
+        val handler = Handler()
+        handler.postDelayed({
```
SaintDubious: THIS! This right here is gold! You are a super hero @nitehu ! I tried this out and it solved my problem of the out-of-order `onCreate`. I'm afraid I didn't have much more time to look at everything else you did, I'll keep looking. You've done a bunch that I don't understand, but I think I've handled some of the parts that are just stubs in your code (the browsing parts). Maybe we can join them up. There might be a solution lurking in here :)

I'm optimistic again. Did you manage to get a working dev environment? You must have to get all this code done. Is it using hardware or did you get it done with an emulator? It's a pain to set up isn't it?

nitehu: Yes I've set up the dev environment based on the link you shared. I searched all around the web but it seems this won't work with an emulator, just with a real device. Of course I hate this solution already, doing all the magic rituals of debug service starting, adb port forward, etc. everytime I want to start testing... XD

I'm currently working on the tricky Now Playing parts, I also think that the browsing is more straightforward and can use the solutions you've already implemented.

nitehu:
MediaSession is now in its own class. It can be accessed easily where it is needed, and it can communicate with events.
I've also fixed a bunch of things on the Now Playing screen, it doesn't blink during track changes, displays the playback progress correctly, etc.

Still must do the display of songs in the queue correctly, and the shuffle and repeat settings. After that only the media browsing remains (which will be easier I hope :)

nitehu:
I've done a few hours of reading, experimenting and testing (and swearing), and it seems the MediaBrowser is entirely incompatible with the nicely implemented ViewModels. Not only it needs a lot of hacking to make it work, the MediaBrowser service doesn't support LiveData, paging, errors or anything advanced... it only supports a static list of items, and nothing else.

So I've tried a different approach. The MediaBrowserService simply retrieves the list of items to display from the subsonic api. Clean and works (and can be adapted easily to the special needs of Android Auto...).
I'm open to suggestions and new ideas, but for me it seems this is the way.

SaintDubious:
Wow! This looks so much better. I can't believe how much progress you've made. Sorry about the swearing part, I too spent many an evening wanting to smash my keyboard. 

It's looking really great, I love the menus, you even found nice icons. Super slick. I can't test playing yet, all the folders are empty for me, is that still a work in progress?

nitehu:
Yep sorry, only the playlists and the artist list is implemented yet. I'll try to add other things in the next few days...

nitehu:
I think every important part is implemented now for Android Auto to include this in a beta release.
Still missing:
- Shuffle and repeat settings for playback, maybe starring
- Little improvements for the media browsing, like "More" and "Refresh" menu items
- Better error handling, etc

For me it generally works well, there are some edge cases which can mess up Android Auto, like changing servers on the phone while it is connected - these can be fixed later, if necessary (one shouldn't play with their phone while driving anyways :)). 
But of course this PR still needs thorough testing.

@SaintDubious thank you very much for experimenting with the different solutions, and generally paving the way! This PR wouldn't have happened without your work.

Would you like to help with testing if you have some free time? Two heads are better than one :)

ultrasonic/src/main/AndroidManifest.xml
nitehu: AutoMediaPlayerService is now removed.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/AutoMediaBrowserService.kt
nitehu: MediaSession handling is now refactored. Starting the MediaPlayerService and LifecycleSupport is still necessary, when Ultrasonic is booted up by Android Auto

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/AutoMediaBrowserService.kt
nitehu: These were removed, can be added back when necessary

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
nitehu: Now it is implemented correctly

SaintDubious:
This is fantastic, so amazed to see it working so well. I love the "Play All" button at the top of song lists. I had struggled to come up with something that could work, I think this is a great solution.

I pushed a small change which is just undoing some of my earlier attempts to reuse the `GenericListModel` stuff.

I can't really be of much help with the code as I no longer understand it, BUT I can always test. I only managed to find one bug, and it's a bit hard to re-create:

- clear out the cache so that Ultrasonic has no song currently playing
- run auto and select an album (I tested this through the album list)
- play an individual song (NOT the Play All button)

When I do this it just shows the spinner and tells me it's fetching my content. Note that if a previous song was playing or if you use the "Play All" button everything works fine.

nitehu:
Nice catch! It is now fixed I hope... ;)

tzugen:
Hey, just wanted to let you know that I am traveling without my laptop, so won't be able to review until August. But don't wait for me if other reviewers find it OK!

nitehu:
@tzugen Oh okay! Sorry to disturb you. Have a nice trip!

nitehu (approved):

tzugen:
Hi!
I am back...
Should I review this?

nitehu:
Yes that would be absolutely great, thank you! However currently Im on a
vacation so no need to rush :)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/di/MediaPlayerModule.kt
```diff
@@ -1,6 +1,5 @@
 package org.moire.ultrasonic.di
 
-import org.koin.android.ext.koin.androidContext
```
tzugen, line 3: Nice :+1:

tzugen (approved):
I couldn't test the actual car features for real, but the code looks fine and I didn't see any regressions in the other parts of the app.

tzugen:
I just went ahead and merged it :balloon: 
If there are follow ups, just open a new PR

PR #559, opened 2021-08-20
Add better offline Support

Maxmystere:
Add visibility of offline supported calls into MainFragment and improved OfflineMusicService->getRandomSongs()


ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/OfflineMusicService.kt
```diff
@@ -257,9 +256,9 @@ class OfflineMusicService : MusicService, KoinComponent {
         if (children.isEmpty()) {
             return result
         }
-        val random = Random()
+        children.shuffle()
         for (i in 0 until size) {
-            val file = children[random.nextInt(children.size)]
+            val file = children[i % children.size]
```
tzugen: As far as I understand this logic,
if we request 40 random songs, but there are only 20 songs in total, 
it will return the 20 songs two times. 

Maybe we should change the 
`for (i in 0 until size)` so that the loop size is never larger than the amount of actual songs.

tzugen (changes requested):
Thanks, looks good!

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/OfflineMusicService.kt
Maxmystere: Yeah it was the previous behavior and i put a comment FIXME about it but CI refused it so i removed it

I will update behavior to better match it

tzugen:
You can use TODO, that is allowed by the CI :)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/OfflineMusicService.kt
tzugen: You can use TODO, that is allowed by the CI :)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/OfflineMusicService.kt
```diff
@@ -257,7 +258,8 @@ class OfflineMusicService : MusicService, KoinComponent {
             return result
         }
         children.shuffle()
-        for (i in 0 until size) {
+        val finalSize: Int = min(children.size, size)
+        for (i in 0 until finalSize) {
```
tzugen: :+1:

tzugen (approved):
Thanks, that is good.
I want to run it on device before I merge,
which will be wednesday proabably!

tzugen:
Is the movie part working for you?
I get a popup "getVideos isn't available in Offline mode"..

PR #567, opened 2021-08-28
Use semantically correct API endpoint when streaming/downloading.

### LINKED ISSUE #257: FR: support download.view
I'm just looking through the code and I see that ultrasonic only sends stream requests to the server.

What I'd like is to have the choice to send download requests when caching.

I'm new to android but I'll try to make a pull based on scrobbling being enabled. 

I.e. if I enable scrobble support then only send download.view requests

tzugen:
Fixes #257


nitehu (approved):
Works good, review is OK.

As I understand, after this modification the "pinned" songs will be downloaded with the download.view, the other songs on the playlist are still cached with the stream.view.

It isn't a big modification to use the strem.view only for the currently playing song, and use download.view for everything else (which would be the truly semantically correct way), but I'm not sure if it is worth it.
The original problem was that the server won't know what songs the user is playing, but this can't be solved anyway (without scrobbling). Once a song has been downloaded, Ultrasonic won't go to the server again for it.

What do you think?

tzugen:
> Works good, review is OK.
> 
> As I understand, after this modification the "pinned" songs will be downloaded with the download.view, the other songs on the playlist are still cached with the stream.view.
> 
> It isn't a big modification to use the strem.view only for the currently playing song, and use download.view for everything else (which would be the truly semantically correct way), but I'm not sure if it is worth it.
> The original problem was that the server won't know what songs the user is playing, but this can't be solved anyway (without scrobbling). Once a song has been downloaded, Ultrasonic won't go to the server again for it.
> 
> What do you think?

Your understanding is correct.

Right now at the point where the download request arrives in our API classes, there is only the boolean "save" argument, so no way to distinguish between the currently playing song and the next. You are right that that is not perfect, but since it won't be perfect without scrobbling, I would jsut leave it like this and not complicate the logic further..

PR #568, opened 2021-08-28
  Rework Downloader

tzugen:
Downloader is now Kotlin.

    Improved Queue handling using Priority Queue
    Parallel downloading
    Cleaner seperation of Download managing and playlist handling.


nitehu (commented):
Nice work!
The review was OK, I want to test this for 1-2 days to see if everything else is fine.

nitehu:
Tested this with pinning and playback on an emulator and real phone too, seems to work fine.

PR #578, opened 2021-09-10
Added link to the wiki for enabling Ultrasonic with Android Auto

nitehu:
Also see ultrasonic#575
When Ultrasonic is not installed from the Play Store, it won't show up until Unknown Sources is enabled in the Android Auto Settings.
I've created a wiki page describing this problem and how to solve it, but I think it is a good idea to make this more visible including it in the readme.


ogarcia:
LGTM

PR #583, opened 2021-09-15
Download metadata

tzugen:
A bunch of changes in the Download sections:

* Creates a Oflline Metadabase, and saves entries into it when files are cached locally. The entries are not yet used in the UI.
* Run the Download code inside our Foreground service. That means we don't need a wakelock anymore.
* Handle the WifiLocks  inside the Download manager class, instead of creating an individual lock per task.
* Run the Download checker only when needed and not continually.
* Convert FileUtils and Constants to Kotlin


nitehu (commented):
Nice! Tried this out and seems to work fine, but I'd like to test this version for a few days more.
Also I'm wondering, what is holding the wifi and wake lock during song pinning, when the MediaPlayer service isn't runnig?

nitehu:
> Also I'm wondering, what is holding the wifi and wake lock during song pinning, when the MediaPlayer service isn't runnig?

Oh I see, it just starts the service if it isn't running. This displays a big empty notification if the playlist is empty.
Ultrasonic could display "Downloading songs in the background" in this case. And the really nice thing to do would be to stop the service after download complete (if there is nothing to play).
I'll try to think up something to solve this. I'm afraid the real problem now is that the service has more separate responsibilities which can run anytime...

nitehu:
I've added a simple check when the notification would be empty to display a message about background download.
Also there was a minor bug with a null check, apparently f.parent can be null...

I'll test this to see if there are more issues. 
@tzugen would you please check if the modification is OK and works for you?

tzugen:
Thanks, your mod was good.
I think the managing of the Playerstate and the notification and the state of the service is still lacking in general.
There was an issue were we discussed this before..
But its super complicated the whole thing. We should work on this at some point.

PR #585, opened 2021-09-17
Added setting to disable Now Playing List sending for incompatible bluetooth devices

### LINKED ISSUE #581: Bluetooth not updating Now Playing information anymore
## Problem description

The latest version does not update the currently playing Artist/Track after the first track is ended or skipped

### Steps to reproduce

Connect to bluetooth device that supports text, allow first track to finish, or skip to next track

## System information

 * **Ultrasonic version**: 2.23
 * **Android version**: 10
 * **Device info**: Samsung S9+

## Additional notes

Version 2.20 doesn't have the issue, nor do earlier versions. I have not checked 2.20 < n < 2.23, so a commit in that range could be responsible

Thank you!

nitehu:
Fixes #581 
There is a new strange behavior with Android Auto and the MediaSession. 
Android Auto can display a Now Playing queue, and to enable this feature we must set the queue contents in the MediaSession.
MediaSession also uses this information to send the queue over bluetooth for AVRCP 1.4 devices.
However, it seems that this breaks compatibility with some AVRCP 1.3 devices, which stop updating the current track. For some devices the solution is to degrade the AVRCP version on the phone, but for others this isn't enough.

For now the best solution seems to be a compatibility switch, which can disable the sending of the Now Playing queue to these devices.

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/Util.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/Util.kt
@@ -1097,6 +1097,13 @@ object Util {
         return preferences.getBoolean(Constants.PREFERENCES_KEY_SEND_BLUETOOTH_ALBUM_ART, true)
     }
 
+    fun getShouldDisableNowPlayingListSending(): Boolean {
+        val preferences = getPreferences()
+        return preferences.getBoolean(
+            Constants.PREFERENCES_KEY_DISABLE_SEND_NOW_PLAYING_LIST, false
+        )
+    }
+
     @JvmStatic
     fun getViewRefreshInterval(): Int {
         val preferences = getPreferences()
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/Util.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/Util.kt
@@ -15,7 +15,6 @@ import android.content.ContentResolver
 import android.content.Context
 import android.content.DialogInterface
 import android.content.Intent
-import android.content.SharedPreferences
 import android.content.pm.PackageManager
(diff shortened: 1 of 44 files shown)


ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/Util.kt
```diff
@@ -1097,6 +1097,13 @@ object Util {
         return preferences.getBoolean(Constants.PREFERENCES_KEY_SEND_BLUETOOTH_ALBUM_ART, true)
     }
 
+    fun getShouldDisableNowPlayingListSending(): Boolean {
+        val preferences = getPreferences()
+        return preferences.getBoolean(
+            Constants.PREFERENCES_KEY_DISABLE_SEND_NOW_PLAYING_LIST, false
+        )
+    }
+
```
tzugen: could be a kotlin property with a getter. but not really necessary ;)

ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/Util.kt
```diff
@@ -1097,6 +1097,13 @@ object Util {
...
     @JvmStatic
     fun getViewRefreshInterval(): Int {
```
tzugen: Same here. There is a nice refactor tool in Android Studio, in the context menu, called "convert to property" or something like this.

tzugen (approved):
Code looks ok. I made two annotations.
But I can't test this, because I don't own any devices like this.

nitehu:
Ok, you were right, those functions are much better as properties, and I started refactoring them... and then my OCD kicked in, and said that Util shouldn't be the class which contains them. So I moved everything setting related from Util to their own class. I think the code is more readable now, and also Util became a bit smaller monstrosity.

PR #593, opened 2021-10-06
Fixed album lists

nitehu:
Fixed a few issues with album lists:
- The controls didn't scale with landscape mode
- The starring of the albums didn't work
- It seems the starred state wasn't even read up from the Subsonic API correctly

Edit: ok I'm starting to get it... the "starred" property of an album on the original Subsonic API isn't documented, probably that's why it was omitted in Ultrasonic. However, every server supports it that I tried, and it is more user friendly to process it if present.

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/AlbumRowAdapter.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/AlbumRowAdapter.kt
@@ -7,15 +7,22 @@
 
 package org.moire.ultrasonic.fragment
 
+import android.content.Context
+import android.graphics.drawable.Drawable
 import android.view.MenuItem
 import android.view.View
 import android.widget.ImageView
 import android.widget.LinearLayout
 import android.widget.TextView
 import androidx.recyclerview.widget.RecyclerView
+import java.lang.Exception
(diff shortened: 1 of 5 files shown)


ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/AlbumRowAdapter.kt
```diff
@@ -87,4 +103,34 @@ class AlbumRowAdapter(
...
+                    )
+                } else {
+                    musicService.unstar(
+                        if (!useId3) entry.id else null,
+                        if (useId3) entry.id else null,
+                        null
+                    )
+                }
+            } catch (exception: Exception) {
+                Timber.e(exception)
```
tzugen: Little tip:
If you write  `catch (all: Exception)`, then you won't need to suppress an error ;)

tzugen (approved):
Thanks!

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/AlbumRowAdapter.kt
nitehu: Oh thanks! I'll try to remember that :)

PR #594, opened 2021-10-07
Added PlaybackComplete intent when a song playback is completed

### LINKED ISSUE #552: Songs not being scrobbled with Simple Scrobbler
## Problem description

I'm using version Simple Scrobbler successfully for ages using various players - all scrobble fine.
Songs I'm listening to with Ultrasonic are not being scrobbled. They **are** being sent with the "now listening" event, but not being submitted once the songs are over.
In the log I see lines such as:

```
I Scrobbler Scrobbled 'submission' for DownloadFile
```

But nothing on Last.fm.
Please note that I'm not trying to scrobble from the server but from within the device itself.

## System information

 * **Ultrasonic version**: *2.22.0*
 * **Android version**: *11*
 * **Device info**: *OnePlus 7T*

nitehu:
Fixes #552 
There was a missing AVRCP intent that should be sent when the playback of a track completes. 
Ultrasonic sent a "stop" intent instead of this, which also caused some strange behavior for connected devices (like blinking stop state)


tzugen (approved):
Code looks good, but I can't test on a real device.
Merge it if you have tested it yourself :)

PR #596, opened 2021-10-08
Added option whether to create a share on the server when sharing songs

### LINKED ISSUE #542: Share song details without creating a share on the server
Most of the time when listening to music with friends, and they ask me "hey, this is a good song, what is it", I don't want / can't give access to my server to them, so it would be nice if I could share with them the details of the song (artist, album, track) in an email or over some instant messenger, so they can search the song on youtube/spotify/bandcamp or wherever later.

Ultrasonic can share a song, but this function currently always creates a share on the server. Then the url of the created share can be sent to others.

Sharing song details without creating a share is a nice feature when
- your server is not configured to allow, or generally doesn't handle shares
- you don't want to give access to your server and/or don't want to maintain the state of shared items on it
- listening in offline mode, or without an internet connection

nitehu:
Closes #542
Added a new checkbox which can be used to turn off share creation on the server. This only works when sharing single songs. In this case only the song details will be shared, for example by using an instant messaging app.

Also added an option to the Player Fragment's menu to share the currently playing song. This can be used to quickly share a nice song with your friends during listening :)

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
@@ -33,6 +33,7 @@ import android.widget.SeekBar
 import android.widget.SeekBar.OnSeekBarChangeListener
 import android.widget.TextView
 import android.widget.ViewFlipper
+import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import androidx.navigation.Navigation
 import com.mobeta.android.dslv.DragSortListView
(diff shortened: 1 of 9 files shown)


ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
```diff
@@ -402,7 +401,7 @@ class PlayerFragment : Fragment(), GestureDetector.OnGestureListener, KoinCompon
                     isVisualizerAvailable = true
                 } else {
                     Timber.d("VisualizerController Observer.onChanged has no controller")
-                    visualizerViewLayout.visibility = View.GONE
+                    visualizerViewLayout.isVisible = false
```
tzugen, line 404: What is the difference between visibility and isVisible?

tzugen: Ok, I see

ultrasonic/src/main/res/xml/settings.xml
```diff
@@ -197,6 +197,11 @@
             a:key="sharingDefaultGreeting"
             a:title="@string/settings.share_greeting_default"
             app:iconSpaceReserved="false"/>
+        <CheckBoxPreference
+            a:defaultValue="true"
+            a:key="sharingCreateOnServer"
+            a:title="@string/share_on_server"
```
tzugen, line 199: I think it would be good to add a `a:summary` to explain what this option means

tzugen (changes requested):

nitehu:
- Added summary
- Reordered sharing settings to be a bit more logical

PR #598, opened 2021-10-09
Update values-zh-rCN/strings.xml

selcarpa:


nitehu:
Thank you very much for the translation!
However, currently this can't be merged, because as per the [contribution guidelines](https://github.com/ultrasonic/ultrasonic/blob/develop/CONTRIBUTING.md), all [commits must be signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification).
Could you please sign the commits?
If you prefer (and allow me) I can push the changes again using my key (but then the commit will have my name on it...)
Also, Ultrasonic has a [Transifex](https://www.transifex.com/ultrasonic/ultrasonic/) page too! I think it's way easier to translate there :)

ogarcia:
The translation cannot be merged even if it is signed. @AethLi you must upload your translation through [Transifex](https://www.transifex.com/ultrasonic/ultrasonic/) because upload it by hand can broke mechanism.
As @nitehu say we could upload your translation file and update strings in Transifex, but it will appear uncredited.

selcarpa:
Understood, I will finish this work later

PR #599, opened 2021-10-10
Moved server selector and settings to the Navigation menu

### LINKED ISSUE #393: Move server selector to menu
Now with the fragments we can improve the visual aspect of application. We can move the server selector buttom from main window to main menu as in Kore app.

Something like this.
Move server selection

We can remove the text "Select server" and only server name remains (like in "_artabria_" sample).

nitehu:
Closes #393 
- The Navigation Menu now displays the selected server name in the header.
- The server name is clickable, and navigates to the server selector fragment (where new servers can be added too). This is now the one central place to manage the server settings and selection.
- Removed the old server selector from the Main fragment
- Removed old server settings from the Settings page
- I had some troubles with the mixed AppCompat and Material themes, so I've updated the old themes to use Material too instead of hacking around it. I think it looks good and everything works where I looked, but needs some more testing to be sure.


ultrasonic/src/main/res/layout/navigation_header.xml
```diff
@@ -1,41 +1,65 @@
...
-            a:layout_height="match_parent"
-            a:gravity="center_vertical"
-            a:paddingStart="16dp"
-            a:paddingLeft="16dp"
-            a:text="@string/common.appname"
-            a:textAppearance="@style/TextAppearance.AppCompat.Title" />
+    <Button
+        a:id="@+id/header_select_server"
+        style="@style/Widget.MaterialComponents.Button.TextButton.Icon"
+        a:layout_width="wrap_content"
```
ogarcia: I think that better is _fill_parent_ with:
```
a:layout_marginEnd="6dp"
a:layout_marginRight="6dp"
a:gravity="center_vertical"
```

ogarcia (commented):
I love it. But the background is some "ugly" now.

What do you think about set the primary color of selected theme as background?

Other idea is use an image, but I don't know which.

ultrasonic/src/main/res/layout/navigation_header.xml
nitehu: Good idea. I've updated the button to cover the navigation bar edge-to-edge, similarly to the other menu items.

nitehu:
Yep I wasn't too fond of that background either.
I've removed it and set the primary color, but it came out a bit too black-and-white. Or did you meant the default, [blueish primary colors of the Material theme](https://material.io/blog/android-material-theme-color)?

ogarcia:
Yes, the  blueish primary color of the Material theme

nitehu:
Oh I totally forgot, the navigation menu had a slightly customized theme that's why it was black and white. I added back the material colors to it.

tzugen:
Let's assign a (user-modifiable) color to each server. Then the user can see in one glance which server is active..

nitehu:
> Let's assign a (user-modifiable) color to each server. Then the user can see in one glance which server is active..

I'd love this feature sooooo much :)
I'll try to add it when I have some time.

tzugen:
let me know when you want it reviewed :sparkle:

nitehu:
I couldn't find more problems, so I think this is ready for review.
Currently it uses the default material colors for the server color until the color is changed in the server settings.

tzugen:
Review ok!

I tweaked some visual things:
* The offline entry has now also a circle around it for consistency
* The selection color is a bit lighter and now comes much closer to the native selection color

I am not sure about the drop shadow at the bottom of the header.
It looks a bit off (on the bottom it is darker than the area that followed) 
I tried to get a native android shadow using elevation, but that didn't work.

ultrasonic/src/main/res/layout/navigation_header.xml
```diff
@@ -1,40 +1,80 @@
...
-            a:text="@string/common.appname"
-            a:textAppearance="@style/TextAppearance.AppCompat.Title" />
+    <ImageView
+        a:id="@+id/img_shadow"
+        a:layout_width="match_parent"
+        a:layout_height="6dp"
+        a:layout_marginTop="-6dp"
+        a:scaleType="fitXY"
+        a:src="@drawable/drop_shadow"
+        app:layout_constraintTop_toBottomOf="@+id/header_select_server" />
```
tzugen: I am not sure about the drop shadow at the bottom of the header.
It looks a bit off (on the bottom it is darker than the area that followed)
I tried to get a native android shadow using elevation, but that didn't work.

nitehu:
Thanks!

I'm not too fond of that shadow either, it can be removed if that's better. The header just looks a bit plain for me with a single color.
I also tried to change the angle of the drop shadow to 45 or 135, that makes it a bit less "shadowy" and more like a divider gradient.

or my other idea is to add a "watermark" effect to the header by replacing the shadow with something like this:

```
   <ImageView
        a:id="@+id/img_backdrop"
        a:layout_width="260dp"
        a:layout_height="260dp"
        a:layout_marginStart="140dp"
        a:layout_marginTop="-110dp"
        a:importantForAccessibility="no"
        a:src="@drawable/ic_stat_ultrasonic"
        a:rotation="330"
        a:alpha="0.2"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toTopOf="parent"
        tools:tint="@color/selected_menu_dark" />
```

what do you guys think?
@ogarcia do you have any better ideas/suggestions?

tzugen:
I actually liked the shadow a bit :)
I checked up a bit, and there seems to be no way to do this with elevation.

I think then if we want to keep it we should tweak it. We should make a version for light and dark theme and the color gradients need to be adjusted a bit.

ogarcia:
Great work @nitehu but as @tzugen say the shadow is very ugly.

I think that we can use a solution like Yatse:
image

image

It uses the lines pattern and apply a color filter with your selection. In offline mode the header is gray

image

I think that any lines pattern works:

image

This idea is the same that we have now, but without center.

ogarcia:
For example we can use this as `ic_header_bg`:
```
<?xml version="1.0" encoding="UTF-8"?>
<svg width="256" height="128" version="1.1" viewBox="0 0 67.733 33.867" xmlns="http://www.w3.org/2000/svg" xmlns:cc="http://creativecommons.org/ns#" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
<metadata>
<rdf:RDF>
<cc:Work rdf:about="">
<dc:format>image/svg+xml</dc:format>
<dc:type rdf:resource="http://purl.org/dc/dcmitype/StillImage"/>
</cc:Work>
</rdf:RDF>
</metadata>
<g transform="translate(0 -268.42)" fill-opacity=".1">
<rect y="268.42" width="67.733" height="33.867" style="paint-order:normal"/>
<path d="m28.575 268.42h39.158v33.867h-67.733z" stroke-opacity=".26667" stroke-width="3.7213"/>
<path d="m39.158 302.29h-39.158v-33.867h67.733z" stroke-opacity=".26667" stroke-width="3.7213"/>
<path d="m19.579 302.29h-19.579v-33.867h33.867z" stroke-opacity=".26667" stroke-width="2.6313"/>
<path d="m48.154 268.42h19.579v33.867h-33.867z" stroke-opacity=".26667" stroke-width="2.6313"/>
</g>
</svg>
```

And, what about remove the logo and Ultrasonic text? Leave only server button. Is better in small screens and looks great.

Note: The `ic_header_bg` is an example. If you have better ideas or are better designers than me, feel free :wink:

tzugen:
I would vote for keeping the ultrasonic text and logo :)

26.10.2021 09:42:36 Óscar García Amor ***@***.***>:

nitehu:
Ok so here are some shots:

This is the original .svg by ogarcia:

This one is tweaked a bit to be brighter at the right like yatse:


These are without logo:



I already like all of them, but I have bad taste for things like UI :)

ogarcia:
The tweaked one looks better than original. I think that without logo it has the advantage of taking up less space, which is good for small screens. For example, in my Samsung J2 I don't need to scroll if I remove the logo.

tzugen:
Is it possible to have a repsonsive layout? Like hide the logo on small screens and show it on large screen?

ogarcia:
IMHO, the logo in the header has no sense now. Before it made sense because there was no element to put there, but now having the server selector we can remove the logo because we have this new element to put in header.

Maxmystere:
I think the good thing with having the logo on big screen is that it will lower all the buttons and allows the navigation menu to be more reachable with one finger
image

nitehu:
I can also add a setting to hide the logo :)

ogarcia:
> I think the good thing with having the logo on big screen is that it will lower all the buttons and allows the navigation menu to be more reachable with one finger image

This depends on size of phone, size of your hand and if you are left or right handed. :wink: 

Looking at other apps that have account button, they don't have logo, for extample: Telegram, Yatse, Element, Boost, OctoDroid, Transdrone, Kubenav, Readably...

> I can also add a setting to hide the logo :)

:joy: :joy: :joy: Nah, is better do the choice and go on :smile:

nitehu:
I've removed the logo and set the background. I also had to change when the foreground switches from dark to light, because the background is now darker. 
If you guys agree we can go with this now at least until we have a better idea what else to put in the header.
Please let me know if this can be merged or if I should change anything else.

tzugen:
I tweaked the layout a bit, to align all text items in a row.

But there is a last minor problem:
In light theme the default violet background has a white text on it, and it looks nice.
But in dark theme the text is suddenly black and has low contrast...

nitehu:
You're right, I've updated the code to compute correctly for the default colors too.

tzugen (approved):

tzugen:
Ok, that works now :) I am happy for you to merge it

ogarcia:
Fine!

In next iterations we can change behabiour to work as Yatse or Boost. Instead of click on server and open _a new window_ to select it, show the server list and _Manage hosts_ button in navigation bar.

What do you think?

tzugen:
Like a drop down menu? I like it

nitehu:
Yep I thought the same, just that change would have been too much for this issue.

PR #600, opened 2021-10-12
Migrate Permission utitlity to Kotlin, increase min SDK to 17

tzugen:
I realized that we relied on READ_EXTERNAL_STORAGE permission, which is actually minimum API 16.
The new API is now 17, because that allowed to cleanup a number of RTL issues from lint.

Contains also some new shortcut to Settings.kt class


nitehu:
> Error: For language "cs" (Czech) the following quantities are not relevant: many [UnusedQuantity]  

[Rant on] Okay so now its the problem of the developer to check the translations of a language, and know what are its rules... I'm starting to hate lint more and more. [Rant off]

nitehu (approved):
Review OK

ogarcia:
The problem with Czech is that in Transifex the language uses the `quantity="many"` so in next language update we will have the same problem. Maybe we must add an exception in lint to ignore this _error_.

PR #602, opened 2021-10-13
Fix NPE

tzugen:
See this bug on G_play:
https://play.google.com/console/u/0/developers/8975783336696288755/app/4975608144956407500/vitals/crashes/444b69da/details?days=30&versionCode=97


nitehu (approved):
Looks good!

PR #603, opened 2021-10-13
Update Gradle and ktlint

tzugen:
This fixes the annyoing ktlin warning:

`Gradle detected a problem with the following location: '/ultrasonic/ultrasonic/src/test/java'. Reason: Task ':ultrasonic:ktlintTestSourceSetFormat' uses this output of task ':ultrasonic:ktlintKotlinScriptFormat' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.`


nitehu (approved):
Review OK

PR #604, opened 2021-10-13
Implement a Download view

tzugen:
`Downloader.kt` now has observable LiveData attributes, which are used by a new `DownloadsFragment.kt` to display a list of currently downloading songs.

The display is working but basic for now (there is no interaction possible, this will be added in a later stage).

Other changes:
- `GenericEntry` now has a mandatory id value
- `GenericEntry` implements a new `Identiafiable` interface
- Removed setData function in the Fragments, and just use the standard `submitList`
- Rename `onDownloadListChanged` to `onPlaylistChanged`
- Add a DownloadStatus enum to `DownloadFile.kt`


ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/DownloadsFragment.kt
```diff
@@ -0,0 +1,225 @@
...
+                holder.status.text = null
+            }
+            DownloadStatus.DOWNLOADING -> {
+                holder.status.text = Util.formatPercentage(downloadFile.progress.value!!)
+                image = SongView.downloadingImage
+            }
+            else -> {
+                holder.status.text = null
+            }
+        }
```
tzugen, line 160: There is some code duplication with SongView.kt here. I will remove it soon, when I migrate the TrackCollection stuff to the new system.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/DownloadsFragment.kt
```diff
@@ -0,0 +1,225 @@
...
+    override fun onItemClick(item: DownloadFile) {
+        // Do nothing
+    }
+
+    override fun setTitle(title: String?) {
+        FragmentTitle.setTitle(this, Util.appContext().getString(R.string.menu_downloads))
+    }
+}
+
+class DownloadRowAdapter(
```
nitehu, line 90: Now I had to check and you're right, the kotlin style guide encourages to put multiple relevant classes in the same file. This looks really strange to me, but OK

nitehu (changes requested):
The code seems to be ok, there are some issues I found otherwise:

- Are you sure you want to close 441 with this? That issue seems to contain other ideas too, like cancelling...
- The Downloads page could write something when there are no downloads instead of being totally empty
- Enqueueing a playlist or an album doesn't automatically start playback as before - something gone wrong detecting when a song is playable

I'll test this some more on the weekend and will see if there is anything else...

Edit: some more findings...
- if I shut down my server during download, the items in the Downloads screen sometimes start to wander up and down. I'm not sure what is the order they should be in, is this intentional?
- I see in the code there are now download statuses for the files, but I can't see it displayed. It would be nice to know why some files remained in the download queue. Also I'm a bit confused when the items just simply disappear after a time when the download fails - maybe it would worth adding them to a limited-time failed list, so the user can see what happened. But then again, maybe these improvements should be added later when the base functions work.

tzugen:
>  Are you sure you want to close 441 with this? That issue seems to contain other ideas too, like cancelling...

You are right. I have remove the link

>   Enqueueing a playlist or an album doesn't automatically start playback as before - something gone wrong detecting when a song is playable

This should be fixed with f50d84f . I don't really understand why calling checkDownloads() at the wrong time interferes with AutoPlay, but so it is..

> if I shut down my server during download, the items in the Downloads screen sometimes start to wander up and down. I'm not sure what is the order they should be in, is this intentional?

See below. I would guess that it is because files are failing, and then are being retried again, but not at the top of the queue, but starting again at the bottom of the queue.

> I see in the code there are now download statuses for the files, but I can't see it displayed. It would be nice to know why some files remained in the download queue. Also I'm a bit confused when the items just simply disappear after a time when the download fails - maybe it would worth adding them to a limited-time failed list, so the user can see what happened. But then again, maybe these improvements should be added later when the base functions work.

This I can explain: Some months ago a retry count was added to the download file. Because a user had high mobile bill because Ultrasonic tried to continously redownload a failed file. Now the reload count is five and after five the download is cancelled.

These two things could be addressed by showing information about the failure status. That would be also nice to see for example in the playlist.

tzugen:
Regarding the second of your points
>  The Downloads page could write something when there are no downloads instead of being totally empty

I am currently working in merging the Track ViewHolder from here with the legacy ViewHolder used in TrackCollectionFragment.

I turned out to be more difficult as expected, because RecyclerView doesn't offer built-in way to handle the selection of elements (unlike ListView did). Therefore I will be using a framework which provides this functionality for RecyclerView (https://github.com/drakeet/MultiType). 

So I would prefer to leave this point open now, because the implementation will change again, and then I don't have to do unnecessary work ;)

core/domain/src/main/kotlin/org/moire/ultrasonic/domain/Identifiable.kt
```diff
@@ -2,9 +2,8 @@ package org.moire.ultrasonic.domain
 
 import androidx.room.Ignore
 
-open class GenericEntry(
-    @Ignore override val id: String
-) : Identifiable {
+abstract class GenericEntry : Identifiable {
```
nitehu, line 5: This generated an exception at Ultrasonic startup, when the playlist is loaded back from file:

W/FileUtil: Failed to deserialize object from /data/user/0/org.moire.ultrasonic.debug/cache/downloadstate.ser
    java.io.InvalidClassException: org.moire.ultrasonic.domain.MusicDirectory$Entry; no valid constructor

...as it needs a parameterless constructor for deserialization.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/Downloader.kt
```diff
@@ -91,10 +92,21 @@ class Downloader(
...
+        if (
+            executorService == null ||
+            executorService!!.isTerminated ||
+            executorService!!.isShutdown
+        ) {
             start()
         } else {
-            executorService?.execute(downloadChecker)
+            try {
+                executorService?.execute(downloadChecker)
```
nitehu, line 103: This exception killed Ultrasonic for me during testing, I think somehow it tried to start while the executor was shutting down.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -159,7 +161,10 @@ class MediaPlayerService : Service() {
     }
 
     fun notifyDownloaderStopped() {
-        stopIfIdle()
+        // TODO It would be nice to know if the service really can be stopped instead of just
+        // checking if it is idle once...
+        val handler = Handler(Looper.getMainLooper())
+        handler.postDelayed({ stopIfIdle() }, 1000)
```
nitehu, line 167: This is the part that caused the autoplay to not work...
CheckDownloadsInternal called stop() in the Downloader, that stopped the MediaPlayerService before it could execute the Play...
Currently I have no better idea than to check this with a bit of a delay.

Also, I think you may check if the checkDownloads calls can be added back, because they won't interfere with the autoplay anymore I hope...

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -740,14 +745,16 @@ class MediaPlayerService : Service() {
...
         private val instanceLock = Any()
 
         @JvmStatic
         fun getInstance(): MediaPlayerService? {
             val context = UApp.applicationContext()
-            synchronized(instanceLock) {
-                for (i in 0..19) {
+            for (i in 0..19) {
+                if (instance != null) return instance
+                synchronized(instanceLock) {
```
nitehu, line 760: This synchronized check also caused a few failed autostarts for me, I think with this fixed ordering it shouldn't fail anymore. (This isn't a new issue, I just found it now...)

nitehu (approved):
I've added a few fixes, could you please check if they are ok?

nitehu:
Oh sorry, I forgot to mention, these fixes were necessary because there were still problems with the autostart.
As the problem is concurrency related it wasn't easy to reproduce... but fairly often, when Ultrasonic coldstarted (e.g. after exit and restart) the autoplay had failed. With these changes this issue seems to be gone (I just hope I haven't added a new one somewhere...)

core/domain/src/main/kotlin/org/moire/ultrasonic/domain/Identifiable.kt
tzugen: Yes, that is a very good fix for the issue, to simply declare the property abstract. Much nicer than having a constructor parameter for an abstract class

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
tzugen: Thanks for finding the root of this. I started to explore it, and then realized, "oh no, there is a race condition somewhere" and once I started adding debug breakpoints the race condition disappeared...

tzugen: The single calls are not needed anymore, i had initially added them out of lazyness to the methods which take care of removing stuff from the playlist, but have replaced them with only the little code that is actually needed in these moments

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
tzugen: Can you explain what the difference makes?

tzugen:
Rebased on develop

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
tzugen: Ok, i get it :).
I added a comment explaining the whole thingy

PR #607, opened 2021-10-18
Use delegate for Settings, reduces Boilerplate code

### LINKED ISSUE #601: Use delegated properties for preference access
Nite has added a nice Settings.kt class.
Using delegated properties we could reduce the boilerplate alot,
eg. https://github.com/mohamad-amin/KPreferences

tzugen:
Closes #601


nitehu (approved):
Review OK

PR #611, opened 2021-10-23
Rename DownloadQueueSerializer to PlaybackstateSerializer

tzugen:
... because it serializes the playlist and not the download queue.

Also replace a Consumer use with a plain lambda.


nitehu (approved):
Review OK

PR #612, opened 2021-10-23
Cleanup a bunch of preference setters and getter by using the new utility class

tzugen:
* Cleanup a bunch of preference setters and getter by using the new utility class
* Migrate SettingsFragment to Kotlin
* Remove Consumer.java class
* Remove leftover build switches from when minSDK was 14 (now Lollipop = 21)


nitehu (approved):
Review OK.
During testing the settings page threw me an exception, because some parameters were not nullable, I've fixed that. All else seems to work OK.

PR #615, opened 2021-10-30
Api30

### LINKED ISSUE #613: targetSdkVersion must be 30 or higher
To upload new releases after november, 1 we must set `targetSdkVersion` to 30 or higher

tzugen:
Experimental. Haven't tested.

* Update API to 30
* Remove dexter
* Remove Filepicker code

Closes #613


ultrasonic/src/main/AndroidManifest.xml
```diff
@@ -29,6 +28,7 @@
         android:label="@string/common.appname"
         android:usesCleartextTraffic="true"
         android:supportsRtl="false"
+        android:preserveLegacyExternalStorage="true"
```
tzugen, line 31: #613

ultrasonic/src/main/kotlin/org/moire/ultrasonic/imageloader/BitmapUtils.kt
```diff
@@ -31,7 +31,7 @@ class BitmapUtils {
             if (entry == null) return null
             val albumArtFile = FileUtil.getAlbumArtFile(entry)
             val bitmap: Bitmap? = null
-            if (albumArtFile != null && albumArtFile.exists()) {
+            if (albumArtFile.exists()) {
```
tzugen: albumArtFile is always non-null

ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/FileUtil.kt
```diff
@@ -475,4 +472,47 @@ object FileUtil {
...
+            val newFile = File(defaultMusicDirectory.path, file.name)
+
+            // Try to rename
+            val success = file.renameTo(newFile)
+            if (success) break
+
+            // If rename failed try to copy and delete
+            var someErrors = false
+            file.copyRecursively(newFile, false) { _, _ ->
+                someErrors = true
```
tzugen: This would currently block the UI!

nitehu:
I've checked this out, the cache folder selection works good, looks good and it was less painful to implement than I thought it would be :)
Of course the returned uri based path can't be used anywhere else in Ultrasonic with the File API... I'm afraid we need to write an abstraction layer for the file operations (similarly to FileUtils) which would use the DocumentsProvider...

tzugen:
> I've checked this out, the cache folder selection works good, looks good and it was less painful to implement than I thought it would be :) Of course the returned uri based path can't be used anywhere else in Ultrasonic with the File API... I'm afraid we need to write an abstraction layer for the file operations (similarly to FileUtils) which would use the DocumentsProvider...

Thanks for having a look!

This is what I feared, that we won't be able to use the URI much. But since I think nearly all file access uses the FileUtil class, we can probably just fix it there and be done.

Reminder to self: move atomic copy function from util to fileutil

tzugen:
I just realized its much larger than expected, because we are used to just handling around a Java File Object, writing to it and extending paths from it, and all that doesn't work anymore.

So we need to search for all usages of java.nio.file and rewrite them :cry: 

Maybe https://github.com/K1rakishou/Fuck-Storage-Access-Framework could help.

I think I won't be able to finish this is the in the foreseable future, because my studies are getting more intense now. I will leave this here, maybe it can be a good starting point for someone else..

nitehu:
Ok no problem! 
I'll pick this up after finishing the work on the current PRs - there is no point to work on anything else since we won't be able to update Ultrasonic in the Play Store without this.
This FSAF library seems good, I thought exactly that someone had to already implement something similar :)

nitehu:
I've killed a few hours into this, but for now I'll put it aside to check the other PRs.

Currently I'm trying to create an abstraction for the file operations which will work with File and the Storage Access Framework too. There are a few use cases which won't work even with FSAF it seems:
- Getting the parent of a file with SAF is not trivial at all. I'm trying to keep track of the parent by manually storing the directory where the file was found/created in a property.
- Naturally paths are completely different in File and SAF, and there's no sure way to translate from one to the other (just maybe some hacks). This will affect the playlist saving mostly, the saved playlists won't be readable by other music players. Then again, as all other players move to SAF too, this won't be much of a problem...
- SAF won't work with non-existent files like File. This means that `File.exists` has no real replacement. In these cases I'm trying to refactor to use a simple string path instead of the File. SAF and File can both check if a string path exists.  

Of course my code is a big mess now, so I can't commit anything yet :(

tzugen:
Yes, thanks for having a stab a this :smiley_cat: 
When I started looking at what needed to be done I got quickly overwhelmed...

nitehu:
Music cache in the Storage Access Framework works now.
It still has a few TODOs and a few bugs here and there, but most importantly it is unusably slow :(
I'm not sure why, I'll try to catch the problem with the profiler...

tzugen:
I heard a few reports that SAF is really slow.

BTW, does the profiler work for you? I get only an overview of how much cpu was used over time, but can't see the footprint of single methods

nitehu:
Never tried it yet :D
I just hoped it would print something useful... but if it doesn't work that way I can still add a few loops and log lines and try to profile by eye ;)

tzugen:
It has a basic and an advanced mode.
In the advanced mode you get this flamechart which shows you what took long time.
The basic mode is just a mem/cpu graph

tzugen:
https://developer.android.com/studio/profile/android-profiler#advanced-profiling

This is the how to how you are supposed to get it working

nitehu:
Eh... the profiler is sh*t, but the logs caught it:
The PlayerFragment and the UpdateView both have scheduled loops executing a few times a second, checking files with SAF. That's what kills the main thread.
I'll try to refactor these to use Rx.

tzugen:
OK, I have a 70% finished PR to remove all UpdateViews.

Maybe you can just comment out the problematic code in the loops and if we then get acceptable performance we can merge this and then I will finish the other PR?

19.11.2021 22:04:12 Nite ***@***.***>:

nitehu:
We don't have to hurry that much, I need a few more days anyway to clean up the TODOs and test everything.
I've optimized my code and commented out the loops, performance seems OK now (for small playlists). Just simply finding a file with SAF takes ~40ms on the emulator. I want to test this on a real device, and also check my playlist with 1000 songs :)
Until everything works fine I'm afraid to merge in anything. I think it is still better to leave the user-selectable cache directory out of the next release if necessary than annoy the users with a bad solution. But I'll leave the UpdateViews to you then.

tzugen:
Sounds like a good plan :)

tzugen:
And sorry if it came across this way, with my previous comment I didn't want to impress hurry, but just tell you that I had been working on a related project, so that we don't do double work!

nitehu:
It's OK I just meant that maybe you'll finish your refactor before I can merge anything :)
And thanks for letting me know, I won't touch anything UpdateView related :D

nitehu:
After quite a few hours of optimizing and bug hunting I'm starting to get slightly optimistic with this PR.
Unfortunately the caching in FSAF wasn't usable for us (or maybe just I wasn't able to figure out how to use it), so I had to add a cache for known existing and non-existing files. Theoretically while Ultrasonic is in the foreground it assumes that the music files are only changed by it, and it can use the caches instead of reading everything again from SAF. I think this only causes problems if someone deletes/renames music files during playback, which is an untested (and maybe unsupported) use case anyways.
But with the caching in place, Ultrasonic seems to be as fast with SAF as with Files, even with big playlists.
I hope the remaining work is just bug hunting and testing, and of course code cleanup where necessary.

nitehu:
Just to see if it is possible, I've removed FSAF entirely. This commit may be rolled back if necessary.
FSAF had two problems:
- the caching still had problems with the latest version too, so our own was still needed
- I'm really not sure why it replaced all spaces with underscores in the filenames. SAF works well with spaces, and the underscores made the offline mode (and the files in the cache) ugly.  

The DocumentFile is an easy, built-in solution, which has some performance problems in itself, but with the caching in place its almost as fast as FSAF was.
We may still try out DocumentsContract, which according the docs has better performance... but that may be another refactor for a later time.

tzugen:
Yes drop it if its not helpful...

Here is another approach to caching:
https://github.com/android/storage-samples/blob/master/ActionOpenDocumentTree/app/src/main/java/com/example/android/ktfiles/CachingDocumentFile.kt#L25

tzugen:
Maybe you saw this:
https://www.reddit.com/r/androiddev/comments/bbejc4/caveats_with_documentfile/

nitehu:
Thanx! I saw that listFiles and findFile were slow as hell, but I didn't imagine getting the name of the file would be a problem too...
Well then, I'll try out DocumentsContract :)

tzugen:
Hi :)

I managed to get the advanced profiler working (using [this trick](https://stackoverflow.com/a/62932163) ).

## Startup
In startup there is a lot of I/O because of the CacheCleaner. The profiler shwos that the listFiles function is the most expensive. As the reddit post i linked also specifically mentions listFiles, i wonder if it would be enough to implement the code linked there (the custom query). Might be easier than the whole Contract thingy.

Bildschirmfoto von 2021-12-11 17-23-43

## Browsing
In browsing the slowest function is actually not listFiles, but the function populateWithDataFrom, which is basically the ID3 parser. In the future I want to solve this with the Room database.

Bildschirmfoto von 2021-12-11 17-29-06

In total the performance felt fine for me!

tzugen:
Here is a link to a DocumentContract equivalent of listFiles: 
https://github.com/davidliu/SAFTraversal/blob/4a6624546236f0d5ab359b19feedd616036d8aaf/app/src/main/java/com/evie/saftraversal/MainActivity.java#L116

tzugen:
But I don't think you need to actually bother. It would be trivial to make the Cache Cleaner run only when all downloads are finished, and not after each download.

nitehu:
Thanks for the profiling!
Actually the cache cleaning runs pretty fast for me, it is also on a background thread so the user won't see it. But of course it would be nice to run it less frequently :)
The real problem now is the handling of very big playlists. When the playlist is loaded (e.g. at startup) Ultrasonic will check the status of every file in it, which means that it will search for a bunch of not-existing files. It is very hard to build a cache for it, if we don't want to read up the whole music cache to memory :)
This isn't that bad, my playlist with ~1500 songs loads in a few seconds. Still, I'll check if it is really necessary to find every file at the start.

tzugen:
Your last commit looks like a very elegant solution.

Regarding the Playlist, where is it happening in the PlaylistDeserializerr? Could we solve it by storing the Playlist in the database?

nitehu:
It seems it is in here:

```
isPathExists:59, Storage (org.moire.ultrasonic.util)
<init>:82, DownloadFile (org.moire.ultrasonic.service)
getDownloadFileForSong:449, Downloader (org.moire.ultrasonic.service)
getDownloadFile:510, Downloader (org.moire.ultrasonic.service)
addToPlaylist:366, Downloader (org.moire.ultrasonic.service)
addToPlaylist:180, MediaPlayerController (org.moire.ultrasonic.service)
restore:77, MediaPlayerController (org.moire.ultrasonic.service)

```

I think it can be solved by making the DownloadFile status lazy, so they are not loaded all at once. Unfortunately this won't be solved by the database, as Ultrasonic needs to check whether the file exists or not, but it would be a great step forward if it would only check those files that it is working with (downloading or displayed in a RecyclerView).
I'll check it when I have more time this week.

nitehu:
I've added the lazy status, so now the loading of big playlists is very fast. We still have a problem when Ultrasonic goes to Offline mode with a big playlist - that will run `clearIncomplete`, which will also check all files on the playlist if they exist or not. We're talking about a few seconds delay and a rare usecase, so I'm not planning to fix this in this issue.
Otherwise I'm using this PR as my daily driver for weeks now without problems, so I think it is ready for review and beta testing.

nitehu (approved):
I've checked the modifications again, and made a minor cleanup.
@tzugen I can't request a review because this is your PR, but please if and when you have free time, help me check and test this :)

tzugen:
Great! I have installed it now and will review the code in the next days.

ultrasonic/lint-baseline.xml
```diff
@@ -41,18 +55,29 @@
...
+    </issue>
+
+    <issue
+        id="PluralsCandidate"
+        message="Formatting %d followed by words (&quot;tracks&quot;): This should probably be a plural rather than a string"
+        errorLine1="    &lt;string name=&quot;select_album.n_unselected&quot;>%d tracks unselected.&lt;/string>"
+        errorLine2="    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
+        <location
+            file="src/main/res/values/strings.xml"
+            line="150"
```
tzugen: Check this

ultrasonic/src/main/java/org/moire/ultrasonic/util/StreamProxy.java
```diff
@@ -155,14 +153,27 @@ boolean processRequest() {
...
-            File file = new File(localPath);
-            if (!file.exists()) {
-                Timber.e("File %s does not exist", localPath);
-                return false;
-            }
+			if (Storage.INSTANCE.isPathExists(localPath)) return true;
 
-            return true;
-        }
+			// Usually the .partial file will be requested here, but sometimes it has already
```
tzugen: @nitehu Your IDE used tabs for indenting here, maybe check its settings.

tzugen: Looks like the file was messed up containing mixed indetation before. I push a fix

ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/Storage.kt
```diff
@@ -0,0 +1,82 @@
...
+            StorageFile(null, documentFile.uri, documentFile.name!!, documentFile.isDirectory)
+        } else {
+            val file = File(Settings.cacheLocation)
+            if (!file.exists()) return null
+            JavaFile(null, file)
+        }
+    }
+}
+
+fun String.isUri(): Boolean {
```
tzugen, line 88: This should be fixed before the next release.

tzugen: It is used only in the settings. Maybe instead just try to parse the string, and catch the exception if it is unparseable?

tzugen (commented):
Review ok

ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/Storage.kt
nitehu: Thx for finding it, I planned to update this.

PR #616, opened 2021-10-31
Remove global scope use

tzugen:
@nitehu We had talked about this before. You had figured out that we should not use viewmodelScope, because any jobs would be killed when the user navigates away from the fragment.

By reading more in the docs, especially [this](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html#coroutine-scope), i figured out how to create a scope that is not bound to the lifecycle, but also not the globalscope, against which one can find many warnings.

The trick is simply to create a scope using `coroutineScope(Dispatchers.IO)`. That's it ;)


nitehu:
Wow, nice find :) 👍

nitehu (approved):
Review OK

PR #620, opened 2021-11-01
CacheCleaner to Kotlin

tzugen:
* CacheCleaner to Kotlin
* Remove custom Pair implementation

This was part of the API30 patch, but I took it apart, since it is not strictly necessary to be included there.


nitehu (approved):
I've fixed a bug where it tried to list files as if they were directories
Otherwise the review is OK

ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/CacheCleaner.kt
```diff
@@ -139,11 +148,11 @@ class CacheCleaner {
         }
 
         private fun getMinimumDelete(files: List<File>): Long {
-            if (files.isEmpty()) {
-                return 0L
-            }
+            if (files.isEmpty()) return 0L
+
```
tzugen, line 152: Is this part of the kotlin styleguide? i learned that braces are preferable, because you are less likely to introduce errors when you add a second statement later.

tzugen: @nitehu

nitehu: Good question! This was debated more times I can count, and there are numerous pros and cons for each approach... The last general consensus I use today is no brackets when it is a simple statement and fits on one line - this makes functions more compact and at the same time readable at a glance.

But it's just me, we can agree to use another coding style :)

PR #626, opened 2021-11-13
Migrate CacheCleaner to Coroutine

tzugen:
AsyncTask is deprecated with API 30.
Therefore migrate the tasks to coroutines


nitehu (approved):
Review OK
Tested it a bit and seems to work fine.

PR #628, opened 2021-11-16
Remove Silentbackground Task

tzugen:
Remove SilentBackgroundTask class, and replace usages with `launch {}`

Also fixes a UI problem, where long song titles overlayed the track number.


ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
```diff
@@ -233,17 +235,11 @@ class PlayerFragment : Fragment(), GestureDetector.OnGestureListener, KoinCompon
...
-                }
-
-                override fun done(result: Void?) {
-                    onCurrentChanged()
-                    onSliderProgressChanged()
-                }
-            }.execute()
+            launch(CommunicationError.getHandler(context)) {
+                mediaPlayerController.previous()
+                onCurrentChanged()
```
nitehu, line 240: Just a note:
I think these can be refactored to use the RxBus observables.
There are already events for when something changes in the player state or track progress.

nitehu (approved):
Review OK

PR #629, opened 2021-11-16
Remove MyViewFlipper.java

tzugen:
It was a workaround for a bug in Android 2, which is fixed in Android 3 and later.


nitehu (approved):
Review OK

PR #632, opened 2021-11-18
Fix empty notification

### LINKED ISSUE #621: Empty notification shouldn't be visible
## Problem description

Sometimes Ultrasonic displays a stuck empty notification.
This shouldn't happen, when the notification would be empty, the service should be stopped and the notification should be removed.

This problem isn't trivial to reproduce, it may be concurrency related.

tzugen:
It looks like it was possible for the notification to end up empty if the download queue was empty and the currentPlaying was null, but somehow the service was still running. 

Instead of showing an empty notification, the service is now stopped.

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
@@ -73,6 +73,10 @@ class MediaPlayerService : Service() {
     private var currentPlayerState: PlayerState? = null
     private var currentTrack: DownloadFile? = null
 
+    private val manager: NotificationManagerCompat by lazy {
+        NotificationManagerCompat.from(this)
+    }
+
     override fun onBind(intent: Intent): IBinder {
         return binder
     }
@@ -484,7 +488,6 @@ class MediaPlayerService : Service() {
             channel.lockscreenVisibility = Notification.VISIBILITY_PUBLIC
             channel.setShowBadge(false)
 
-            val manager = getSystemService(NOTIFICATION_SERVICE) as NotificationManager
             manager.createNotificationChannel(channel)
         }
     }
@@ -492,15 +495,15 @@ class MediaPlayerService : Service() {
     fun updateNotification(playerState: PlayerState, currentPlaying: DownloadFile?) {
         val notification = buildForegroundNotification(playerState, currentPlaying)
 
+        if (notification == null) {
+            manager.cancelAll()
+            stopIfIdle()
+            return
+        }
+
         if (Settings.isNotificationEnabled) {
             if (isInForeground) {
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                    val manager = getSystemService(NOTIFICATION_SERVICE) as NotificationManager
-                    manager.notify(NOTIFICATION_ID, notification)
-                } else {
-                    val manager = NotificationManagerCompat.from(this)
-                    manager.notify(NOTIFICATION_ID, notification)
-                }
+                manager.notify(NOTIFICATION_ID, notification)
                 Timber.v("Updated notification")
             } else {
                 startForeground(NOTIFICATION_ID, notification)
@@ -517,7 +520,7 @@ class MediaPlayerService : Service() {
     private fun buildForegroundNotification(
         playerState: PlayerState,
         currentPlaying: DownloadFile?
-    ): Notification {
+    ): Notification? {
 
         // Init
         val context = applicationContext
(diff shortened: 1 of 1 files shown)


tzugen:
Should fix #621 

I can test that the normal notification is still working fine, but I can't test whether the bug is definitively fixed.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/MediaPlayerService.kt
```diff
@@ -484,23 +488,22 @@ class MediaPlayerService : Service() {
...
+
         if (Settings.isNotificationEnabled) {
             if (isInForeground) {
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                    val manager = getSystemService(NOTIFICATION_SERVICE) as NotificationManager
-                    manager.notify(NOTIFICATION_ID, notification)
-                } else {
-                    val manager = NotificationManagerCompat.from(this)
-                    manager.notify(NOTIFICATION_ID, notification)
-                }
```
tzugen, line 503: NotificationCompat works across all APIs, no need to add this check here

tzugen:
I am not sure which branch my phone is on atm, but today after a reboot I got an "empty" notification saying: "Downloading media in the background". Directly after staring the phone, before i had opened any apps.

tzugen:
That fixes it.
But also now the service which is being started on boot is killed after 5 seconds.
I think someone should check if the Auto stuff it then still working

hmueller01:
I saw those empty notification on my Pixel 6 Android 12 device using release 3.0.0 as well e.g. after reboot. Just merged this PR into current develop and at least after reboot no incorrect notification.

hmueller01:
After using it 3 days it seems to work perfect. I think it should be merged. Tx. Holger

tzugen:
But do you have android auto? I was not completely sure if it still works, and i ride a :bike:  :smile:

hmueller01:
No, I don't have that.

hmueller01:
It looks like no one uses Android Auto. Maybe it should be merged and go public. If someone finds a problem and opens a new issue it can be fixed / undone then. At the moment it is more annoying having this "permanent" notification (it always shows up after some time without rebooting on my device ...).

hmueller01:
Ups, there is an Android Auto User, see #687. Maybe @Break-Off can test this pull ...

Break-Off:
> Ups, there is an Android Auto User, see #687. Maybe @Break-Off can test this pull ...

I'm wiling to. Please advise on how to to test it. I'm not familiar with compiling for Android.

hmueller01:
If you like I can send you a build apk. But you have to uninstall Ultrasonic first, as it is not signed.

Break-Off:
Sure. What should I look for when testing the build?

hmueller01:
I build a current version (3.1.1) with this patch included and uploaded it to Google drive for testing. 
https://drive.google.com/file/d/1ebEJXuHlKEUSCDgfC4vN7xHv_GL5cZQJ/view?usp=sharing

hmueller01:
What I found with this patch implemented is that when I press Play in the Now Playing view with an empty playlist Ultrasonic crashes with this stacktrace:
```
Android API level: 31
Ultrasonic version name: 3.1.1
Ultrasonic version code: 102


android.app.ForegroundServiceDidNotStartInTimeException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{dfb7e26 u0 org.moire.ultrasonic.debug/org.moire.ultrasonic.service.MediaPlayerService}
	at android.app.ActivityThread.throwRemoteServiceException(ActivityThread.java:1923)
	at android.app.ActivityThread.access$2700(ActivityThread.java:247)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2148)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7839)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)

```
Is this reproducible?

tzugen:
Closing for now, as I will concentrate on moving the playback to media3, and that will provide and handle the notification on its own.

hmueller01:
Without this patch notification does not work correctly on Android 12 (and I assume Android 11 as well). I haven't noticed this as I was using this patch from the beginning. After lates updates it produced some conflicts and I was too lazy refactoring this. Now, I just did and notification works again on my bleeding edge Android ;-)
So it really makes sense to move forward to media3 #724 ...

tzugen:
I have a basic working branch with media3 since yesterday :)
But needs more cleanup before I publish...

hmueller01:
@tzugen You are really the best!!!

PR #642, opened 2021-12-09
Remove MergeAdapter and SackOfViewsAdapter

tzugen:
After the big refactor suroounding UpdateView there were two adapters left.
Theses adapters were now only used in the MainFragment, where the task they were solving (showing a  static menu) could me achieved much less complicated.


nitehu (approved):
Review OK

PR #651, opened 2021-12-20
Updated custom cache location handling to remove isUri

### LINKED ISSUE #645: Fix String.isUri
Checking for a colon seems oversimplified. Ensure there are no weird situations where the check might fail.

nitehu:
Fixes #645 

There was another problem which caused an exception: when the cache location was changed, old downloadFile paths may still  remained in the downloadFileCache.


nitehu:
Eh I'll create this on a branch

PR #652, opened 2021-12-20
Updated custom cache location handling to remove isUri

### LINKED ISSUE #645: Fix String.isUri

nitehu:
Fixes #645

There was another problem which caused an exception: when the cache location was changed, old downloadFile paths may still remained in the downloadFileCache.

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SettingsFragment.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SettingsFragment.kt
@@ -38,7 +38,6 @@ import org.moire.ultrasonic.service.MediaPlayerController
 import org.moire.ultrasonic.service.RxBus
 import org.moire.ultrasonic.util.Constants
 import org.moire.ultrasonic.util.ErrorDialog
-import org.moire.ultrasonic.util.FileUtil.defaultMusicDirectory
 import org.moire.ultrasonic.util.FileUtil.ultrasonicDirectory
 import org.moire.ultrasonic.util.InfoDialog
 import org.moire.ultrasonic.util.MediaSessionHandler
@@ -287,7 +286,7 @@ class SettingsFragment :
         val intent = Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
 
         if (Settings.cacheLocationUri != "" && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-            intent.putExtra(DocumentsContract.EXTRA_INITIAL_URI, defaultMusicDirectory.path)
+            intent.putExtra(DocumentsContract.EXTRA_INITIAL_URI, Settings.cacheLocationUri)
         }
 
         intent.addFlags(RW_FLAG)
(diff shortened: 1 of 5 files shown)


ultrasonic/src/main/kotlin/org/moire/ultrasonic/util/Storage.kt
```diff
@@ -70,22 +71,16 @@ object Storage {
     }
 
     private fun getRoot(): AbstractFile? {
```
tzugen, line 73: <s>It looks like the function is now always returning a non-null value</s>

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/Downloader.kt
```diff
@@ -44,6 +44,7 @@ class Downloader(
 
     private val jukeboxMediaPlayer: JukeboxMediaPlayer by inject()
 
+    // TODO is this cache necessary?
     private val downloadFileCache = LRUCache<MusicDirectory.Entry, DownloadFile>(100)
```
tzugen, line 48: I would say yes. It helps us to avoid creating duplicate DownloadFile instances when showing Entries.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SettingsFragment.kt
```diff
@@ -259,34 +268,32 @@ class SettingsFragment :
...
-                if (!isDefault && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                    intent.putExtra(DocumentsContract.EXTRA_INITIAL_URI, defaultMusicDirectory.path)
-                }
+    private fun selectCacheLocation() {
+        // Choose a directory using the system's file picker.
+        val intent = Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
 
-                intent.addFlags(RW_FLAG)
-                intent.addFlags(PERSISTABLE_FLAG)
+        if (Settings.cacheLocationUri != "" && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
tzugen, line 288: I am not sure about this logic. Shouldn't we always try to set the initial directory, either to the currently picked location or to the default location?

Now it actually sets the initial directory to be the default only if the user has chosen a non default?!

tzugen (changes requested):

ultrasonic/src/main/kotlin/org/moire/ultrasonic/service/Downloader.kt
nitehu: Oh okay, I see now. I thought it was just there to speed things up. I'm also not sure what its worth for playlists with a few hundred - 1k items... Anyways, I've added your comment.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/SettingsFragment.kt
nitehu: Argh, thx for finding this. I've read what I wanted to see and not that was written there.
Of course the logic is meant to be that if there is an already selected uri, show that as the default location.
(I'm not even sure what this does with a file path, maybe nothing)

tzugen (approved):

tzugen:
ok!

PR #653, opened 2021-12-20
Update OkHttp to 4.9.1

tzugen:
We were very behind on OkHttp versions because we needed to support low APIs. We can upgrade now that the minApi is higher


ogarcia:
@tzugen whenever you can create an issue that describes this, categorize it and put in the milestone instead of the pull request :wink: 

Update: I think that is related to #654

nitehu (approved):
Review OK.
I've tested it and seems to work fine.

PR #655, opened 2021-12-21
Create a correctly typed list from getMusicDirectory()

### LINKED ISSUE #650: Release next version
I suggest bumping the version code to 3, as we now have quit a lot of major changes :)
There are a few small issues and PRs left and then it should be ready for beta.

tzugen:
Fixes #650

### DIFF
--- a/core/domain/src/main/kotlin/org/moire/ultrasonic/domain/MusicDirectory.kt
+++ b/core/domain/src/main/kotlin/org/moire/ultrasonic/domain/MusicDirectory.kt
@@ -39,15 +39,15 @@ class MusicDirectory : ArrayList<MusicDirectory.Child>() {
         abstract var album: String?
         abstract var title: String?
         abstract override val name: String?
-        abstract val discNumber: Int?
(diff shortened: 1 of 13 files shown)


ogarcia:
It seems work weel, but I found other bug. In ramdom list (Albums -> Random), after you start scrolling down, the list reloads.

Update: Other bug in playlists. If I open a playlist cannot scroll down the entire list (only few lines)

Update 2: The bug in playlist seems that is in scroller, is bigger than list size:
photo_2021-12-21_15-15-41

The list only have 6 songs but scroller bar is like a big list

tzugen:
Good catch!

tzugen:
I am opening a new PR for 1.
Regarding 2, the scrollbar height is hardcoded in the library: https://github.com/timusus/RecyclerView-FastScroll/blob/69130ef9e59939918e3c7fb1116ec3a050305be1/recyclerview-fastscroll/src/main/java/com/simplecityapps/recyclerview_fastscroll/views/FastScroller.java#L114

3, I can't reproduce...

PR #656, opened 2021-12-21
Don't clear random list when appending

tzugen:


ogarcia:
Related to #650

ogarcia (approved):
It works fine. I think that can be merged :wink:

PR #659, opened 2021-12-24
More tiny and pretty ic_empty icon

ogarcia:


ogarcia:
Is a stupid change, but I think that looks better

tzugen:
Can you give a screenshot?

ogarcia:
Yezzz!
photo_2021-12-24_19-46-49

tzugen:
The icon is OK, but if we use it in this big size, the line weight should be thinner in my opinion...

ogarcia:
> The icon is OK, but if we use it in this big size, the line weight should be thinner in my opinion...

Ok. For now we leave this and I take note to make it more thin in future.

PR #660, opened 2021-12-24
Fix database migrations

ogarcia:
I think that this is better than lost your settings in a downgrade :muscle:


ogarcia:
I will kill the fucking checkstyle and analysis :angry:

tzugen:
> I will kill the fucking checkstyle and analysis 😠

Don't do it! It means us no evil 🐇

tzugen:
The code looks fine. I can't test it in the next days. But I am fine with merging it if you are sure it works good.

tzugen (approved):

PR #667, opened 2022-01-13
Deactivate custom color and set DB to version 3

tzugen:


nitehu (approved):
The text was still visible :Đ
Now it seems to work correctly (but please review my changes).
The DB was also downgraded to v3 without any issue on my emulator :)

nitehu:
Argh! Something happened to my signature! I'll try to fix that asap...
Edit: Ok it works now...

PR #684, opened 2022-02-13
Redesigned the About fragment - removed WebView

### LINKED ISSUE #683: Rewrite the about and remove the webview
Recently our _dear_ friends at Google block the latest update (3.0.1) because they don't like that there is a donate button in the about section and send this:
org moire ultrasonic-InAppExperience-46

But wait, they have another hilarious argument for blocking the app and it's caused by this flow (I am not even going to evaluate it):
org moire ultrasonic-InAppExperience-327

I am using all the donations that have been received to pay for the moire.org domain and they are not enough to pay for anything else. With this I do not mean that I want more money to be donated, on the contrary, the domain is paid until next year and there are funds to pay at least one more year.

Side note for @tzugen @nitehu and rest of contributors. If at any time we need to set up a VPS for testing purposes, these funds can be used.

At this point we have two options.

- The first is what I propose in https://github.com/ultrasonic/ultrasonic/discussions/682 (which may be too radical, but if we all agree, I have nothing more to add).
- The second one is remove the about text and webview and set a simple fixed about window that links to https://ultrasonic.github.io/ and here (https://github.com/ultrasonic/ultrasonic) and add the donation button in Github.

nitehu:
Closes #683 
The About fragment now contains a simple text, and a few buttons which will open links in the default browser on the phone.


nitehu:
I've created the about text by copying it from the original and from github.
@ogarcia would you please check if this was the text and buttons you imagined? Should we change/add/remove anything else?

ogarcia:
I love it. Much better than old webview about. I think that is perfect. I will merge it now and this afternoon I do a new release 3.1.0 with this change and #678

Issue #688, opened 2022-02-16
Connection failure


fiskhandlarn:
## Problem description

I'm getting "Connection failure" as of latest version (3.1.0). I haven't changed any of the settings, I just installed the v3.1.0 apk over the previous version.

### Steps to reproduce

Se above.

## System information

 * **Ultrasonic version**: 3.1.0
 * **Android version**: 11
 * **Device info**: Samsung Z Flip3 5G

nitehu:
Please check your server settings, did they change? There were some related db changes, and maybe there was a bug in them.
Can you add a new server and see if that helps?

fiskhandlarn:
@nitehu Nothing has changed on the server. I played songs from the server yesterday evening with v3.0.1 (or if it was 3.0.0). The only thing that has changed is upgrading Ultrasonic to 3.1.0.

> Can you add a new server and see if that helps?

I only have one server.
I just tried with http://demo.subsonic.org/index.view though (username "guest2", password "guest") and get the same error.

nitehu:
Thanks!
Could you please also check if there is an exception file in the Ultrasonic folder, in the debug log (it can be enabled from the settings) or in logcat?

fiskhandlarn:
@nitehu No exception file, but I enabled the log file:

```log
14:25:30.393: I Storage StorageFile caches were reset 
14:25:34.233: D NavigationActivity Navigated to org.moire.ultrasonic:id/serverSelectorFragment 
14:25:34.234: D ActiveServerProvider Cache is invalidated 
14:25:34.236: D ServerSettingsModel$updateItem updateItem updated server setting: ServerSetting(id=2, index=2, name=demo, url=http://demo.subsonic.org/, color=null, userName=guest2, password=guest, jukeboxByDefault=false, allowSelfSignedCertificate=false, ldapSupport=false, musicFolderId=null, minimumApiVersion=null, chatSupport=null, bookmarkSupport=null, shareSupport=null, podcastSupport=null) 
14:25:34.255: D ActiveServerProvider$getActiveServer getActiveServer retrieved from DataBase, id: 2 cachedServer: ServerSetting(id=2, index=2, name=demo, url=http://demo.subsonic.org/, color=null, userName=guest2, password=guest, jukeboxByDefault=false, allowSelfSignedCertificate=false, ldapSupport=false, musicFolderId=null, minimumApiVersion=null, chatSupport=null, bookmarkSupport=null, shareSupport=null, podcastSupport=null) 
[...]

PR #689, opened 2022-02-16
Updated Gradle to fix R8 bug with Kotlin 1.6

nitehu:
This relates to the issue found in #688, in the release version of 3.1.0.
The problem was that minify/R8 removed some metadata which made the deserialization of responses fail.

I was in a bit of a panic, because this was an obscure error, but the fix seems to be simple enough: the new gradle version contains a fixed R8 which works correctly with Kotlin 1.6.

@ogarcia if you have time, could you please release a patched version? This works now with a release build on my pc, but I want to test a real release which was built by the CI too...

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -3,7 +3,7 @@
 gradle                 = "7.3.2"
 
 navigation             = "2.3.5"
-gradlePlugin           = "7.0.4"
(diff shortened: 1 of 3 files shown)


ogarcia:
@nitehu I'm working now in the release

ogarcia:
All done! Thanks @nitehu

PR #699, opened 2022-03-05
Updated references to context_menu_track to correctly identify events

### LINKED ISSUE #698: Track based context menus do not function correctly in most fragments
## Problem description

Context menus under track based fragments (except Search) don't perform any actions.

### Steps to reproduce

Navigate to Media Library and select any Artist and Album
Long press to bring up context menu
Select play now or any other action and nothing will occur

## System information

 * **Ultrasonic version**: Self built from develop branch
 * **Android version**: Android 12
 * **Device info**: Samsung S10

## Additional notes

I've been looking at making some adjustments to the bookmark fragment for personal use and noticed that there appears to be some inconsistency between the ids under the xml defined layout and the event handler in the fragment file. 

https://github.com/ultrasonic/ultrasonic/blob/82d90a6aee972d5d02b222354c20d66261498d7b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/adapters/TrackViewBinder.kt#L28

https://github.com/ultrasonic/ultrasonic/blob/82d90a6aee972d5d02b222354c20d66261498d7b/ultrasonic/src/main/res/menu/context_menu_track.xml#L5-L23

https://github.com/ultrasonic/ultrasonic/blob/82d90a6aee972d5d02b222354c20d66261498d7b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/TrackCollectionFragment.kt#L633-L679

omgcem:
Additionally updated event handler to correctly play, queue, pin and
download songs now that events are triggering correctly.

Fixes issue #698

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/TrackCollectionFragment.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/TrackCollectionFragment.kt
@@ -709,7 +709,7 @@ open class TrackCollectionFragment : MultiListFragment<MusicDirectory.Child>() {
     }
 
     internal fun getClickedSong(item: MusicDirectory.Child): List<MusicDirectory.Entry> {
-        //This can probably be done better
+        // This can probably be done better
         return viewAdapter.getCurrentList().mapNotNull {
             if (it is MusicDirectory.Entry && (it.id == item.id))
                 it
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/TrackCollectionFragment.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/TrackCollectionFragment.kt
@@ -46,6 +46,7 @@ import org.moire.ultrasonic.util.Constants
 import org.moire.ultrasonic.util.EntryByDiscAndTrackComparator
 import org.moire.ultrasonic.util.Settings
 import org.moire.ultrasonic.util.Util
+import org.moire.ultrasonic.util.Util.toast
 
 /**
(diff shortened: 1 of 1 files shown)


nitehu (changes requested):
Thank you for the fix!
I've tested it, and the menu now seems to work correctly.

It would be a bit nicer if you would reuse the already existing functions (the ones setupButtons() is already calling) in the class, e.g. unpin() for Unpin (it just needs a parameter), and for the other menu items too. In Pin for example downloadBackground() already shows a toast, so showing one there isn't necessary.

What do you think?

omgcem:
I'm happy to have another look at this, I just picked up the new handler from the SearchFragment as that seemed to have been done around the same time as the references were changed.

My only concern would be that the functions under setupButtons() are designed to work around the checkboxed track items and so they contain a function to get the selected songs:
https://github.com/ultrasonic/ultrasonic/blob/7dd479c0d298d7356cc20375636571fe245eaace/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/TrackCollectionFragment.kt#L414-L415

If we added a parameter to unpin, as you suggest, so that we could use it on the context menu then we would also need to amend these existing calls to pass the selected songs as a parameter instead?

Let me know if I'm understanding your request correctly, I've not worked with Kotlin before and only looked at this repo for a few days so I'm still trying to figure it all out.

nitehu:
> If we added a parameter to unpin, as you suggest, so that we could use it on the context menu then we would also need to amend these existing calls to pass the selected songs as a parameter instead?

Yep exactly, so the other call would become unpin(getSelectedSongs())

nitehu (approved):
Review OK!

PR #703, opened 2022-03-11
Translation update

hmueller01:
Updated German translation
Minor adjustments in common strings.xml

### DIFF
--- a/ultrasonic/src/main/res/values-de/strings.xml
+++ b/ultrasonic/src/main/res/values-de/strings.xml
@@ -30,7 +30,7 @@
     <string name="chat.send_a_message">Nachricht senden</string>
     <string name="common.album">Album</string>
     <string name="common.appname">Ultrasonic</string>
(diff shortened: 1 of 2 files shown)


ogarcia:
@hmueller01 first for all, thanks for your contrubution, but there is some changes needed.

* The first one is that te commits must be signed (see [here][1]).
* The second is that the translations are managed in [Transifex][2] so please, send only your proposal of changes in the `strings.xml` file and, once aceppted, you will be able to translate it in Transifex.

[1]: https://github.com/ultrasonic/ultrasonic/blob/develop/CONTRIBUTING.md
[2]: https://www.transifex.com/ultrasonic/ultrasonic/

hmueller01:
Damn, should have read the [CONTRIBUTING.md](https://github.com/ultrasonic/ultrasonic/blob/develop/CONTRIBUTING.md) first ...
Ok, I'll do that stuff.

PR #706, opened 2022-03-12
fixed output of debug file size (in MB)

hmueller01:
This replaces #704 as I have no idea how to sign a commit afterwards.


nitehu (approved):
Thank you for the fix!
Review OK
Tested and the file size now displays correctly.

PR #710, opened 2022-03-18
Add detailed song information to Now Playing

hmueller01:
This is a proposal to show more detailed information to Now Playing (like genre, year, bitrate, file format). It can be switched on/off in the settings. Another idea might be to add this to the menu. But this is not yet implemented.

### DIFF
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
@@ -992,17 +992,20 @@ class PlayerFragment :
                 )
                     currentSong!!.suffix
                 else
-                    String.format(Locale.ROOT, "%s > %s", currentSong!!.suffix,
-                        currentSong!!.transcodedSuffix)
+                    String.format(
+                        Locale.ROOT, "%s > %s", currentSong!!.suffix,
+                        currentSong!!.transcodedSuffix
+                    )
             val details: String = String.format(
                 Util.appContext().getString(R.string.song_details_nowplaying),
                 currentSong!!.genre, currentSong!!.year, currentSong!!.bitRate, fileFormat
             )
             detailsTextView.text = details
-            if (Settings.showNowPlayingDetails)
-                detailsTextView.visibility = View.VISIBLE
-            else
-                detailsTextView.visibility = View.GONE
+            detailsTextView.visibility =
+                if (Settings.showNowPlayingDetails)
+                    View.VISIBLE
+                else
+                    View.GONE
 
             downloadTrackTextView.text = trackFormat
             downloadTotalDurationTextView.text = duration
--- a/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
+++ b/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
@@ -138,8 +138,9 @@ class PlayerFragment :
     private lateinit var playlistFlipper: ViewFlipper
     private lateinit var emptyTextView: TextView
     private lateinit var songTitleTextView: TextView
-    private lateinit var albumTextView: TextView
     private lateinit var artistTextView: TextView
+    private lateinit var albumTextView: TextView
+    private lateinit var genreTextView: TextView
     private lateinit var detailsTextView: TextView
     private lateinit var albumArtImageView: ImageView
     private lateinit var playlistView: RecyclerView
@@ -176,9 +177,10 @@ class PlayerFragment :
         playlistFlipper = view.findViewById(R.id.current_playing_playlist_flipper)
         emptyTextView = view.findViewById(R.id.playlist_empty)
         songTitleTextView = view.findViewById(R.id.current_playing_song)
-        albumTextView = view.findViewById(R.id.current_playing_album)
         artistTextView = view.findViewById(R.id.current_playing_artist)
-        detailsTextView = view.findViewById(R.id.current_playing_details)
+        albumTextView = view.findViewById(R.id.current_playing_album)
+        genreTextView = view.findViewById(R.id.current_playing_genre)
+        detailsTextView = view.findViewById(R.id.current_file_details)
         albumArtImageView = view.findViewById(R.id.current_playing_album_art_image)
         positionTextView = view.findViewById(R.id.current_playing_position)
         downloadTrackTextView = view.findViewById(R.id.current_playing_track)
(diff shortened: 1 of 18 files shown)


ogarcia:
@hmueller01 can you include a screenshot of your proposal?

hmueller01:
These checks are really bad. Androidstudio did the optimization of the imports and this tools complains ...
```
/home/circleci/ultrasonic/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt:10:1 Imports must be ordered in lexicographic order without any empty lines in-between
/home/circleci/ultrasonic/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt:17:1 Wildcard import (cannot be auto-corrected)
/home/circleci/ultrasonic/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt:20:1 Wildcard import (cannot be auto-corrected)
/home/circleci/ultrasonic/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt:57:1 Wildcard import (cannot be auto-corrected)
/home/circleci/ultrasonic/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt:63:1 Wildcard import (cannot be auto-corrected)
```

hmueller01:
> @hmueller01 can you include a screenshot of your proposal?
Shure. :-)

Screenshot_20220319-105711~2

ogarcia:
IMHO, I think that displaying the information as it appears in the screenshot completely breaks the style of the application. In addition, it should be taken into account that many users do not have the correct tags in their files, so it is possible that this line appears blank (the genre and the year).

The idea is good, but maybe it would be more interesting to show some kind of button (an i of information for example) and that when pressing it all the information appears. I don't know, it would be necessary to think about it more in detail.

And about CI and style, using asterisks for imports is not allowed because you lose detail of all the functions you use, that's why it fails in that step. Sometimes these tests can be frustrating but they help us to make good code. :wink:

tzugen:
> These checks are really bad. Androidstudio did the optimization of the imports and this tools complains ...
> 
> ```
> /home/circleci/ultrasonic/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt:10:1 Imports must be ordered in lexicographic order without any empty lines in-between
> /home/circleci/ultrasonic/ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt:17:1 Wildcard import (cannot be auto-corrected)
> ```

Yes, it's annoying. If you find a way to fix it its very welcome!

ogarcia:
> Yes, it's annoying. If you find a way to fix it its very welcome!

For me these both errors make a lot of sense. Have imports in lexicographic order helps to locate it, and wildcard imports must be avoided in any language (because you either import too much or you don't know what you import). :wink:

tzugen:
The error is good, but it is annoying that Android Studios Formatter actually sometimes creates these issues.
But you can configure it in the settings. I have opened a PR to include Android Studios code formatting settings into VCS => #711

tzugen:
> IMHO, I think that displaying the information as it appears in the screenshot completely breaks the style of the application. In addition, it should be taken into account that many users do not have the correct tags in their files, so it is possible that this line appears blank (the genre and the year).

Also text should be right aligned like the rest of the text. I think if it is blank, the line can be made to collapse.

hmueller01:
> IMHO, I think that displaying the information as it appears in the screenshot completely breaks the style of the application. In addition, it should be taken into account that many users do not have the correct tags in their files, so it is possible that this line appears blank (the genre and the year).

Ok, too bad that you do not like it. Is there a way to not break the style of your application? This was the way I like it. And not showing these information was on mayor thing I did not like in Ultrasonic. But yes, you are right the code is not finished in case these data are not available. I do not have such music. Have to prepare a special song ...

> The idea is good, but maybe it would be more interesting to show some kind of button (an i of information for example) and that when pressing it all the information appears. I don't know, it would be necessary to think about it more in detail.

No, I do like the idea of a button which I have to press every time. First this information is hidden in the default case. And second I was proposing a menu entry to enable/disable this quickly.

hmueller01:
Another proposal for these information could be
`<left> genre                <center> year                    <right> bitrate and file format.`
That would better fill the line. I tried the left aligned version before and war not so happy with that way.

tzugen:
> Another proposal for these information could be `<left> genre <center> year <right> bitrate and file format.` That would better fill the line. I tried the left aligned version before and war not so happy with that way.

Could also work:

## Basic display (current situation)
```
Title
Artist                Tracknumber
Album                 Duration
```

## Advanced display

```
Title
Artist                 Tracknumber
Album (Year)           Duration
Genre                  Bitrate/Format
```

ogarcia:
I think that @tzugen idea can work well

hmueller01:
Ok, if I find time I will try to implement this, as I like it too. I will show the result here for further voting. What do you think about the menu entry?

ogarcia:
> Ok, if I find time I will try to implement this, as I like it too. I will show the result here for further voting. What do you think about the menu entry?

I think that the menu entry is unnecesary in this context because all appearance settings are controlled in application settings.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
```diff
@@ -979,8 +984,41 @@ class PlayerFragment :
         if (currentPlaying != null) {
             currentSong = currentPlaying!!.song
             songTitleTextView.text = currentSong!!.title
-            albumTextView.text = currentSong!!.album
             artistTextView.text = currentSong!!.artist
+            albumTextView.text = currentSong!!.album
+            if (currentSong!!.year != null && Settings.showNowPlayingDetails)
+                albumTextView.append(String.format(Locale.ROOT, " (%d)", currentSong!!.year))
+
+            genreTextView.text = currentSong!!.genre
```
ogarcia: All this block should be inside the `if (Settings.showNowPlayingDetails)`, mostly because if the user is not showing this information it is useless to calculate it.

ultrasonic/src/main/res/values/strings.xml
```diff
@@ -332,6 +334,7 @@
     <string name="settings.wifi_required_title">Download on Wi-Fi only</string>
     <string name="song_details.all">%1$s%2$s</string>
     <string name="song_details.kbps">%d kbps</string>
+    <string name="song_details.nowplaying">%1$s&#160; • &#160;%2$d&#160; • &#160;%3$d kbps %4$s</string>
```
ogarcia: I think that this is useless now, can be deleted.

ultrasonic/src/main/res/values/strings.xml
```diff
@@ -332,6 +334,7 @@
     <string name="settings.wifi_required_title">Download on Wi-Fi only</string>
     <string name="song_details.all">%1$s%2$s</string>
```
ogarcia: Please, set this string as **translatable="false"**:
```
<string name="song_details.all" translatable="false">%1$s%2$s</string>
```

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
```diff
@@ -174,8 +177,10 @@ class PlayerFragment :
         playlistFlipper = view.findViewById(R.id.current_playing_playlist_flipper)
         emptyTextView = view.findViewById(R.id.playlist_empty)
         songTitleTextView = view.findViewById(R.id.current_playing_song)
-        albumTextView = view.findViewById(R.id.current_playing_album)
         artistTextView = view.findViewById(R.id.current_playing_artist)
+        albumTextView = view.findViewById(R.id.current_playing_album)
+        genreTextView = view.findViewById(R.id.current_playing_genre)
+        detailsTextView = view.findViewById(R.id.current_file_details)
```
ogarcia: I prefer:
```
bitrateFormatTextView = view.findViewById(R.id.current_playing_bitrate_format)
```

It would be necessary to change the name of the variables everywhere obviously. :wink:

ogarcia (changes requested):
There are a couple of details that need to be changed to make it a little clearer. Remember that it must pass the CI tests in order to be merged.

ultrasonic/src/main/res/values/strings.xml
hmueller01: Ups, forget to delete that. Thanks for watching!

ultrasonic/src/main/res/values/strings.xml
hmueller01: Ok, done.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
hmueller01: Ok, it is calculated just once per song so it does not really matter. But I changed it. Done.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
hmueller01: Refactoring makes the job, so thats easy. ;-) Done.

hmueller01:
This is how it looks right now:
Screenshot_20220321-175951

hmueller01:
@ogarcia should I remove `<string name="song_details.all">%1$s%2$s</string>` from all language files to resolve this?
```
/home/circleci/ultrasonic/ultrasonic/src/main/res/values-it/strings.xml:290: Error: The resource string "song_details_all" has been marked as translatable="false" elsewhere (usually in the values folder), but is translated to "it" (Italian) here [Untranslatable]
    <string name="song_details.all">%1$s%2$s</string>
```

ogarcia:
> @ogarcia should I remove `<string name="song_details.all">%1$s%2$s</string>` from all language files to resolve this?

Yes, please.

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
```diff
@@ -979,8 +982,34 @@ class PlayerFragment :
...
             songTitleTextView.text = currentSong!!.title
-            albumTextView.text = currentSong!!.album
             artistTextView.text = currentSong!!.artist
+            albumTextView.text = currentSong!!.album
+            if (currentSong!!.year != null && Settings.showNowPlayingDetails)
+                albumTextView.append(String.format(Locale.ROOT, " (%d)", currentSong!!.year))
+
+            if (Settings.showNowPlayingDetails) {
+                genreTextView.text = currentSong!!.genre
+                genreTextView.visibility =
```
tzugen: Shorter syntax:  .isVisible = true/false

tzugen (changes requested):

tzugen:
A little tip, you can run the CI commands also locally. It's usually quicker than to push the commit and wait. The exact commands can be found in the .Circleci folder

ultrasonic/src/main/kotlin/org/moire/ultrasonic/fragment/PlayerFragment.kt
hmueller01: Tx for the hint! Done.

ogarcia (approved):

tzugen (approved):

PR #711, opened 2022-03-19
Don't exclude the whole .idea folder

tzugen:
Inside the idea folders are config files which are useful to be checked in to vcs.
Such as the fomarting configurations regarding imports:

https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,15 @@ captures/
 
 # Intellij
 *.iml
-.idea/
+.idea/libraries
+.idea/dictionaries
+.idea/workspace.xml
+.idea/shelf
+.idea/jarRepositories.xml
+.idea/gradle.xml
+.idea/navEditor.xml
+.idea/modules.xml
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -37,7 +37,15 @@ captures/
 
 # Intellij
 *.iml
-.idea/
```
ogarcia, line 40: I think it is better to keep the whole .idea directory ignored, just in case a new file appears that we do not have control over. Since you have added the interesting files to the version control, they will be maintained, so there will be no problem.

What do you think?

ogarcia (changes requested):

.gitignore
tzugen: Yes :). I have not thought about that, but that should work. I'll check it out.

tzugen:
Merged with your proposed changes

PR #712, opened 2022-03-19
Add locale chooser

### LINKED ISSUE #669: Feature request: option to change language
The translations are very different in terms of quantity and quality.
Furthermore, in version 3, the gender stars have found their way into the German translation. Regardless of what I think of them, they are very inconsistent. Sometimes they are present and at other places they are not present.
I think these are enough reasons to integrate an option for language selection independent of the set system language.

tzugen:
Adds an option to change to language manually.
Fixes #669


ultrasonic/build.gradle
```diff
@@ -14,10 +14,11 @@ android {
 
         minSdkVersion versions.minSdk
         targetSdkVersion versions.targetSdk
-        resConfigs 'cs', 'de', 'en', 'es', 'fr', 'hu', 'it', 'nl', 'pl', 'pt', 'pt-rBR', 'ru', 'zh-rCN', 'zh-rTW'
```
tzugen, line 17: I have removed this because it is annoying to maintain and I guess we want to include all languages which we have files for anyway...

ultrasonic/build.gradle
```diff
@@ -14,10 +14,11 @@ android {
...
 
     }
 
+    bundle.language.enableSplit = false
+
```
tzugen, line 21: Otherwise Play Store might deliver a locallized stripped version, which will break the dynamic language chooser.
The additional languages are not more than a few thousand bytes anyway

ultrasonic/src/main/res/xml/settings.xml
```diff
@@ -343,16 +349,14 @@
             a:summary="@string/settings.hide_media_summary"
             a:title="@string/settings.hide_media_title"
             app:iconSpaceReserved="false"/>
-    </PreferenceCategory>
-    <PreferenceCategory
-        a:title="@string/settings.features_title"
-        app:iconSpaceReserved="false">
```
tzugen, line 349: I have removed the one item category "Features" because the star preference can be better grouped somewhere else.

tzugen:
I have added a new language variant, German (traditional spelling). In this variant the contested strings have male gender. I think this should be acceptable for all now.

ogarcia:
@tzugen we cannot add traditional german variant because is not supported by transifex :disappointed:

ultrasonic/src/main/res/values/arrays.xml
```diff
@@ -236,5 +236,38 @@
         <item>@string/settings.playback.bluetooth_a2dp</item>
         <item>@string/settings.playback.bluetooth_disabled</item>
     </string-array>
+    <string-array name="languageNames" translatable="true">
```
ogarcia: I think that this should be `translatable="false"`, if not, it will appear to translate the texts `@string/language.default`, `@string/language.zh`, etc...

ultrasonic/build.gradle
ogarcia: I think that is better maintain it becase we need to add languages to language array (and the language name) and this is a good method to remember it.

ultrasonic/src/main/res/values/arrays.xml
```diff
@@ -236,5 +236,40 @@
...
+        <item>en</item>
+        <item>fr</item>
+        <item>de</item>
+        <item>de-DE-1901</item>
+        <item>hu</item>
+        <item>it</item>
+        <item>es</item>
+        <item>pl</item>
+        <item>pt</item>
+        <item>pt-BR</item>
```
ogarcia, line 272: The language code is `pt-rBR`, isn't it?

ultrasonic/src/main/res/values/arrays.xml
```diff
@@ -236,5 +236,40 @@
...
+        <item>@string/language.pl</item>
+        <item>@string/language.pt</item>
+        <item>@string/language.pt_br</item>
+        <item>@string/language.ru</item>
+    </string-array>
 
+    <string-array name="languageValues" translatable="false">
+        <!--suppress CheckTagEmptyBody -->
+        <item></item>
+        <item>zh</item>
```
ogarcia: We have two chinesse language Chinese (China) `zh-rCN` and Chinese (Taiwan) `zh-rTW`

ultrasonic/src/main/res/values/arrays.xml
```diff
@@ -236,5 +236,40 @@
         <item>@string/settings.playback.bluetooth_a2dp</item>
         <item>@string/settings.playback.bluetooth_disabled</item>
     </string-array>
+    <string-array name="languageNames" translatable="true">
+        <item>@string/language.default</item>
+        <item>@string/language.zh</item>
```
ogarcia: As we have two chinnese language I think that is better to use `zh_cn` Chinese (China) and `zh_tw` Chinese (Taiwan)

ogarcia (changes requested):
In general terms, the change is very good. The only problem is that we can't put the traditional German because Transifex doesn't support it (that change would have to be removed). In any case, if the German translation is consistent and the screen reader has no problems with the "gender star", I don't think it is necessary to add another translation. :wink:

ultrasonic/src/main/res/values/arrays.xml
tzugen: The IETF BCP 47 language tag is `pt-BR`, and that is what `java.util.Locale` understands. Don't ask me why in the folder names Android invented their own standard.

ultrasonic/src/main/res/values/arrays.xml
tzugen: ok!

tzugen:
Ok! I have removed it for now and differentiated between Taiwan & China.

ogarcia (approved):

ultrasonic/src/main/res/values/strings.xml
```diff
@@ -96,6 +96,21 @@
     <string name="error.label">Error</string>
     <string name="jukebox.is_default">Jukebox By Default</string>
     <string name="lyrics.nomatch">No lyrics found</string>
+    <string name="language.default">Default</string>
```
ogarcia: I think that better is **System default** instead of simply **Default**:
```
<string name="language.system">System default</string>
```

ogarcia (changes requested):
It works great :heart:

The only thing I would change is use **System default** for default language option (I think it is clearer) and it would be ready to merge.

ogarcia (approved):

hmueller01:
I found an issue with this in Settings.kt. The shareGreeting String is created with the wrong language / locale setting (I switched to English and it stays German). I think it is because the context is copied before the locale is changed. But I have no idea how to change this ...
```
    val shareGreeting: String?
        get() {
            val preferences = preferences
            val context = Util.appContext()
            val defaultVal = String.format(
                context.resources.getString(R.string.share_default_greeting),
                context.resources.getString(R.string.common_appname)
            )
```
`context.resources.configuration.locales` returns here `[de_DE,en_DE]` (which is my default locale).
Simple solution would be not to use this String.format here and to write "Ultrasonic" in R.string.share_default_greeting (as it was done on some other places). But there are more places where `UApp.applicationContext().getString()` is used ...
So maybe you have a better idea.

Issue #718, opened 2022-03-27
Add ID3 support when offline


tzugen:
This is a tracking bug.

I have finished a working code, where Ultrasonic uses the Room DB to display the offline cached songs.
It is faster than walking through the File tree and parsing everything manually all the time.

Before this feature can be merged, some preliminary changes are still needed.
I will put them in seperate PRs to keep everything manegable.

### Steps:

- [x] 1.  #722 #719
- [x] 2. Wait one minor release
- [x] 3. #720 
- [ ] 4. Add Album and Track entities to the DB, Make PrimaryKey of Artist, Album and Track a composite key referencing the server id as well as the items id. Add a preference to use ID3 when offline: #750

tzugen:
We are moving to Gitlab. Please continue the discussion there: https://gitlab.com/ultrasonic/ultrasonic/-/issues/718

PR #720, opened 2022-03-27
Don't manage ID manually, but use autoGenerate to ensure uniqueness

tzugen:
See #718  This is step 3

Make server id unique. Needed to reference the servers from which tracks or albums are cached precisely.


tzugen:
Defacto merged already through #726

PR #729, opened 2022-04-07
made button off mode better visible

hmueller01:


hmueller01:
debug build fails. What's wrong with that?

tzugen:
Merged

PR #731, opened 2022-04-08
Target SDK 31

hmueller01:
make build working with target sdk 31


hmueller01:
Ok, now it passes all checks.
The include/exclude xml's are still empty templates. AFAIK this backups everything. And this is different from old Manifest.xml setting. Any suggestions what to exclude?
BTW why was it set to false? Backing up the settings would make sense ...

ultrasonic/src/main/AndroidManifest.xml
```diff
@@ -63,15 +66,16 @@
...
+            android:exported="false">
 
             <intent-filter>
                 <action android:name="androidx.media3.session.MediaLibraryService" />
                 <action android:name="android.media.browse.MediaBrowserService" />
             </intent-filter>
         </service>
 
-        <receiver android:name=".receiver.UltrasonicIntentReceiver">
+        <receiver android:name=".receiver.UltrasonicIntentReceiver"
```
tzugen, line 77: I thinks this needs to be exported, otherwise other apps won't see it.

ultrasonic/src/main/AndroidManifest.xml
```diff
@@ -63,15 +66,16 @@
 
         <service android:name=".playback.PlaybackService"
             android:label="@string/common.appname"
-            android:exported="true">
+            android:exported="false">
```
tzugen, line 70: I am unsure if it needs to be exported

tzugen:
> Ok, now it passes all checks. The include/exclude xml's are still empty templates. AFAIK this backups everything. And this is different from old Manifest.xml setting. Any suggestions what to exclude? BTW why was it set to false? Backing up the settings would make sense ...

We should backup settings but not the media files IMO.

Regarding the flag, I think you can take out the switch based on build version. The flag has been around for a while, we can just add it by default. Devices with API lower than 23 will just ignore it.

hmueller01:
Current min API is 21. If I leave off the if clause we get this error:

`Field requires API level 23 (current min is 21): android.app.PendingIntent#FLAG_IMMUTABLE`

hmueller01:
> We should backup settings but not the media files IMO.

Agreed. Can you help me, what I need to specify in the exclude tag?

tzugen:
> Current min API is 21. If I leave off the if clause we get this error:
> 
> `Field requires API level 23 (current min is 21): android.app.PendingIntent#FLAG_IMMUTABLE`

OK, then I made a wrong assumption :)

tzugen:
<exclude domain="file" path="."/>
<include domain="sharedpref" path="."/>
<include domain="database" path="."/>

tzugen:
```
<exclude domain="file" path="."/>
<include domain="sharedpref" path="."/>
<include domain="database" path="."/>
```

This is with what I would start.

See https://developer.android.com/guide/topics/data/autobackup#Files


In the other file I would use the same definition in the cloud backup part, and allow all files in the device to device transfer part

hmueller01:
I wasn't sure if this `path="."` works or if I need to find out the real names of e.g. the database. So just tried your suggestion.

BTW: `<exclude domain="file" path="."/>` did not work (path "." is not valid). Using `root` instead.

tzugen (changes requested):
Looks good except the exported=false changes (see comments)

tzugen:
I have merged the changes with the immutability flag. The other things we need to research a bit more i think.

tzugen:
@hmueller01 Hm, this got automatically closed somehow and I cant reopen it :(
Can you try?

tzugen:
I think you need to click edit and change the target branch to 'develop'

ogarcia:
I can't reopen it either. Let's see if @hmueller01 can look at it and if he can't either, the best thing to do is to open a new PR.

hmueller01:
Couldn't reopen this. Was easier to create a new PR #754.

PR #737, opened 2022-05-01
Support HTTP basic authentication

### LINKED ISSUE #351: "Invalid URL" when trying to use HTTP basic auth
## Problem description

I have a Subsonic API that uses normal auth and HTTP basic auth on top of the Subsonic auth. 

FWIW this is navidrome with nginx in front of it. I have a user account set up on navidrome and nginx basuc auth http.

I believe it should be possible to do this "double authentication" (with nginx and navidrome) but when I try and enter the following a URL of the form `https://user:password@host.com` I get the error `Please specify a valid URL`.

### Steps to reproduce

Try and add a Server with URL `https://user:password@host.com`

## System information

 * **Ultrasonic version**: 2.15.0
 * **Android version**: 10
 * **Device info**: NA

## Additional notes

It's possible I've misunderstood something here - but looking at the API for the web browser client I user it looks like this should work (i.e. the double auth).

i-do-cpp:
You can now specify HTTP basic authentication credentials in the server URL, e.g. `https://myuser:mypassword@music.server.com`.

Tested on https://github.com/navidrome/navidrome/releases/tag/v0.47.5 behind Nginx with basic HTTP authentication enabled.

closes #351

### DIFF
--- a/core/subsonic-api/src/main/kotlin/org/moire/ultrasonic/api/subsonic/SubsonicAPIClient.kt
+++ b/core/subsonic-api/src/main/kotlin/org/moire/ultrasonic/api/subsonic/SubsonicAPIClient.kt
@@ -8,6 +8,7 @@ import java.security.cert.X509Certificate
 import java.util.concurrent.TimeUnit.MILLISECONDS
 import javax.net.ssl.SSLContext
 import javax.net.ssl.X509TrustManager
+import okhttp3.Credentials
 import okhttp3.OkHttpClient
 import okhttp3.ResponseBody
 import okhttp3.logging.HttpLoggingInterceptor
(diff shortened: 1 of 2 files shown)


nitehu:
Thank you very much for your work!
Your code seems to be fine, unfortunately I can't test that the double authentication works, but I'll execute some regression tests to see if everything is OK and merge this.

core/subsonic-api/src/main/kotlin/org/moire/ultrasonic/api/subsonic/SubsonicAPIClient.kt
```diff
@@ -73,7 +74,19 @@ class SubsonicAPIClient(
...
+            if (originalRequest.url.username.isNotEmpty() &&
+                originalRequest.url.password.isNotEmpty()
+            ) {
+                newRequestBuilder = newRequestBuilder.addHeader(
+                    "Authorization",
+                    Credentials.basic(
+                        originalRequest.url.username,
+                        originalRequest.url.password
+                    )
+                )
```
tzugen: Why instantiate two Builders after each other?
Just remove the build() call from the first builder, reuse it in your logic,
and then call build() in the end.

tzugen (changes requested):
Thanks! Also please check if this works in the media3 branch :)

core/subsonic-api/src/main/kotlin/org/moire/ultrasonic/api/subsonic/SubsonicAPIClient.kt
i-do-cpp: Can you elaborate which builders you mean? In this interceptor we first use an `HttpUrl.Builder` and then a `Request.Builder`

tzugen: You are totally right, and quickly reading I didn't see the difference :)
But the reassignment on Line 81 should not be needed. The reason that these methods on the builder returns a Builder ist to enable chaining. The instance of the builder that you call the method on is the same instance that is being returned.

i-do-cpp: Thank you. I removed the assignment

tzugen (approved):

nitehu (approved):
Review OK

PR #750, opened 2022-06-16
Finish offline ID3 support

tzugen:
See #718 for more info.

Add Album and Track entities to the DB, Make PrimaryKey of Artist, Album and Track a composite key referencing the server id as well as the items id. Add a preference to use ID3 when offline.

**Attention:**
If you enable this setting, only those albums or tracks which have been cached by the new method will be shown..

### DIFF
--- a/core/domain/src/main/kotlin/org/moire/ultrasonic/domain/Album.kt
+++ b/core/domain/src/main/kotlin/org/moire/ultrasonic/domain/Album.kt
@@ -1,12 +1,22 @@
+/*
+ * Album.kt
+ * Copyright (C) 2009-2022 Ultrasonic developers
+ *
+ * Distributed under terms of the GNU GPLv3 license.
+ */
+
 package org.moire.ultrasonic.domain
 
+import androidx.room.ColumnInfo
 import androidx.room.Entity
 import androidx.room.PrimaryKey
 import java.util.Date
 
-@Entity(tableName = "albums")
+@Entity(tableName = "albums", primaryKeys = ["id", "serverId"])
 data class Album(
-    @PrimaryKey override var id: String,
+    override var id: String,
+    @ColumnInfo(defaultValue = "-1")
+    override var serverId: Int = -1,
     override var parent: String? = null,
     override var album: String? = null,
     override var title: String? = null,
--- a/core/domain/src/main/kotlin/org/moire/ultrasonic/domain/Album.kt
+++ b/core/domain/src/main/kotlin/org/moire/ultrasonic/domain/Album.kt
@@ -1,8 +1,10 @@
 package org.moire.ultrasonic.domain
 
+import androidx.room.Entity
 import androidx.room.PrimaryKey
 import java.util.Date
 
+@Entity(tableName = "albums")
 data class Album(
     @PrimaryKey override var id: String,
     override var parent: String? = null,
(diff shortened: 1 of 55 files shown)


tzugen:
Don't merge yet. The Ids of Tracks and Albums should become compound ids consisting of server if + entity I'd in order to uniquely reference tracks from different servers

ogarcia:
Initially it looks good, I'm going to leave it running all this week to see if it gives me any strange failure.

tzugen:
@ogarcia 
I have fixed the issue with the artist picture, maybec check that it works now

ogarcia:
Maybe this feature should be dependent on `settings.use_id3`, i.e. you have to have `settings.use_id3` enabled to allow you to enable this. The main reason is that if you have this feature enabled and `settings.use_id3` disabled you are still browsing using ID3 tags.

In other hand, if I have this feature enabled and try to browse offline I always have the list empty. I do setup in develop branch and download some songs, switch to this branch and then try.

tzugen:
I thought I made it dependent, need to check..
Anyway the list is probably empty because I bumped the MetaDB version number, because I added the compound ids and the migration clears the db.

tzugen:
Ok, I made it dependent now :)

ogarcia:
> Ok, I made it dependent now :)

Ok. But has not feedback on setting. You can still enable or disable it. I think that may work as `settings.show_artist_picture` and appear as not selectable.

ogarcia:
> Anyway the list is probably empty because I bumped the MetaDB version number, because I added the compound ids and the migration clears the db.

Yes, but it doesn't matter how many times it goes from online to offline. The list is always empty. It seems that if you upgrade from the develop branch to this one the list is never updated.

tzugen:
> > Ok, I made it dependent now :)
> 
> Ok. But has not feedback on setting. You can still enable or disable it. I think that may work as `settings.show_artist_picture` and appear as not selectable.

Done!

tzugen:
> > Anyway the list is probably empty because I bumped the MetaDB version number, because I added the compound ids and the migration clears the db.
> 
> Yes, but it doesn't matter how many times it goes from online to offline. The list is always empty. It seems that if you upgrade from the develop branch to this one the list is never updated.

I can't reproduce it. Can you uninstall and reinstall?

ogarcia:
> I can't reproduce it. Can you uninstall and reinstall?

Done, same error.

With Android 9 (in emulator for example):
1. Delete App
2. Install develop branch
3. Download a song
4. Stop app, upgrade to this branch, start app
5. Navigate to same song online (it appears as downloaded)
6. Go offline. List is empty.

I think that the problem is with songs that you have already downloaded

tzugen:
You have not read the warning in the first post:

**Attention:**
If you enable this setting, only those albums or tracks which have been cached by the new method will be shown..

This is unavoidable. We don't have the metadata for tracks that have been downloaded before, especially not the ids that the server assigned to them...

ogarcia:
> You have not read the warning in the first post:
> 
> **Attention:** If you enable this setting, only those albums or tracks which have been cached by the new method will be shown..
> 
> This is unavoidable. We don't have the metadata for tracks that have been downloaded before, especially not the ids that the server assigned to them...

:facepalm: Damn right. I hadn't seen it.

I would do the following (to warn users):
1. This new option to be disabled by default.
2. When the user enables it (only the first time) that a dialog box appears with an OK button that warns the user of this behavior.

What do you think?

tzugen:
I have added more clear and colored warning. I didn't feel like a dialog but it should do ;)

tzugen:
Also you can look at albums now in offline mode

Issue #752, opened 2022-06-17
Target SDK 31


ogarcia:
See #731

hmueller01:
I think this can be closed, as we set Target SDK 33 already ...
See #764 , #755

PR #754, opened 2022-06-17
Target sdk 31 into develop

hmueller01:
see Target SDK 31 #731
see #752

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -1,5 +1,5 @@
 ext.versions = [
         minSdk               : 21,
-        targetSdk            : 30,
+        targetSdk            : 31,
         compileSdk           : 31,
 ]
--- a/ultrasonic/src/main/AndroidManifest.xml
+++ b/ultrasonic/src/main/AndroidManifest.xml
@@ -21,6 +21,8 @@
 
     <application
         android:allowBackup="false"
+        android:fullBackupContent="@xml/backup_descriptor"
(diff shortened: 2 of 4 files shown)


ogarcia:
@hmueller01 In order to merge this you will have to make a new branch and put the changes into a single signed commit. Otherwise it will not pass the checks and will not allow us to merge it.

tzugen:
I have just done it.

tzugen (approved):
LGTM

tzugen:
@hmueller01 Something is wrong with your git signing....

hmueller01:
Thanks @tzugen for signing for me. I was sure I set the signoff config in this project. 

`git config format.signoff true`

But somehow it was gone. Hope next time it works ...

We now have 3 errors in AndroidStudio
```
Exported service does not require permission :68
Exported receiver does not require permission :78
Exported content providers can provide access to potentially sensitive data :154
```

> Exported content providers can provide access to potentially sensitive data 
> Inspection info:Content providers are exported by default and any application on the system can potentially use them to read and write data. If the content provider provides access to sensitive data, it should be protected by specifying export=false in the manifest or by protecting it with a permission that can be granted to other applications.


Should we add 

`tools:ignore="ExportedContentProvider"` ...

to these exports and take the risk?

tzugen:
We can just leave it for now. The situation is unchanged to before just with 31 this warning has been introduced..

PR #755, opened 2022-06-17
update build tools

hmueller01:
push to latest gradle version, set targetSdk to 33

Damn it. This was hard to get a signed commit. I had to do
```
git config commit.gpgsign true
git config user.signingkey XYZ
```
Hope this works now. :-)


hmueller01:
Using compileSdk : 33 is not yet recommended ...

> We recommend using a newer Android Gradle plugin to use compileSdk = 33
> 
> This Android Gradle plugin (7.2.1) was tested up to compileSdk = 32
> 
> This warning can be suppressed by adding
>     android.suppressUnsupportedCompileSdk=33
> to this project's gradle.properties
>

hmueller01:
Hm, this is bad. Lint fails with an Error: Library lint checks out of date.
I do not know how to fix this ...

tzugen:
I would say to stay at 31, it is already pretty pretty new, no need to update to some bleeding edge version were the toolset isnt working well

hmueller01:
This is strange. lintRelease is successful on my PC.

> ./gradlew -Pqc lintRelease
> Configuration on demand is an incubating feature.
> Type-safe dependency accessors is an incubating feature.
> 
> BUILD SUCCESSFUL in 26s
> 69 actionable tasks: 4 executed, 65 up-to-date
> 

But it fails on circleci. I updated the android image. But in one of the first runs I saw the use of SDK 30.0.3 (really OLD) and not SDK 33.0. Is there a way to set the Tools SDK? I haven't found something ...

tzugen:
Thi sis the error:

/home/circleci/ultrasonic/ultrasonic/lint-baseline.xml: Information: 1 errors/warnings were listed in the baseline file (lint-baseline.xml) but not found in the project; perhaps they have been fixed? Another possible explanation is that lint recently stopped analyzing (and including results from) dependent projects by default. You can turn this back on with android.lintOptions.checkDependencies=true. Unmatched issue types: ObsoleteLintCustomCheck [LintBaseline]
/home/circleci/ultrasonic/ultrasonic/lint-baseline.xml: Information: 66 errors were filtered out because they are listed in the baseline file, lint-baseline.xml
 [LintBaseline]
/home/circleci/.gradle/caches/transforms-3/f32db2cfde2f54f7eff1514962814390/transformed/jetified-annotation-experimental-1.2.0/jars/lint.jar: Error: Library lint checks out of date.

Lint found an issue registry (androidx.annotation.experimental.lint.ExperimentalIssueRegistry)
which was compiled against an older version of lint
than this one.

This often works just fine, but some basic verification
shows that the lint check jar references (for example)
the following API which is no longer valid in this
version of lint:
com.android.tools.lint.client.api.AnnotationLookup: org.jetbrains.uast.UAnnotation findRealAnnotation(com.intellij.psi.PsiAnnotation,com.intellij.psi.PsiClass,org.jetbrains.uast.UElement)
(Referenced from androidx/annotation/experimental/lint/ExperimentalDetector.class)

Recompile the checks against the latest version, or if
this is a check bundled with a third-party library, see
if there is a more recent version available.

Version of Lint API this lint check is using is 11.
The Lint API version currently running is 12 (7.2). [ObsoleteLintCustomCheck]

   Explanation for issues of type "ObsoleteLintCustomCheck":
   Lint can be extended with "custom checks": additional checks implemented by
   developers and libraries to for example enforce specific API usages
   required by a library or a company coding style guideline.

   The Lint APIs are not yet stable, so these checks may either cause a
   performance degradation, or stop working, or provide wrong results.

   This warning flags custom lint checks that are found to be using obsolete
   APIs and will need to be updated to run in the current lint environment.

   It may also flag issues found to be using a newer version of the API,
   meaning that you need to use a newer version of lint (or Android Studio or
   Gradle plugin etc) to work with these checks.

1 errors, 0 warnings (66 errors filtered by baseline lint-baseline.xml)

tzugen:
The error mentions the annotation-experimental library.

Run `./gradlew ultrasonic:dependencies` to look at the dependencie tree.

Then you can find where this library is imported.

Then update the respective dependency

ultrasonic/lint-baseline.xml
```diff
@@ -103,6 +110,39 @@
...
+    <issue
+        id="ObsoleteSdkInt"
+        message="Unnecessary; SDK_INT is always >= 21"
+        errorLine1="    tools:targetApi=&quot;lollipop&quot;>"
+        errorLine2="    ~~~~~~~~~~~~~~~~~~~~~~~~~~">
+        <location
+            file="src/main/res/drawable/default_ripple.xml"
+            line="6"
+            column="5"/>
+    </issue>
```
tzugen: These don't need to be hidden, they can be fixed ;)

tzugen (changes requested):
lint passes now.
maybe check whats going on in select_ripple.xml and then we can merge

ultrasonic/lint-baseline.xml
hmueller01: True. Done. :-)

hmueller01:
Tx for the help @tzugen !

tzugen (approved):

tzugen:
Merged.

PR #758, opened 2022-06-19
Upgrade to media3-beta01

Maxmystere:
media3-beta01 released recently so I wanted to help upgrade it


hmueller01:
Did a quick checkout to this PR. LGTM. Pixel 6 / Android 12. Tx! :-)
But you still have to fix the failed checks to make it mergeable ...

ultrasonic/src/main/kotlin/org/moire/ultrasonic/playback/PlaybackService.kt
```diff
@@ -48,18 +47,6 @@ class PlaybackService : MediaLibraryService(), KoinComponent {
...
-            session: MediaSession,
-            controller: MediaSession.ControllerInfo,
-            mediaItem: MediaItem
-        ): MediaItem {
-            // Again, set the Uri, so that it will get a LocalConfiguration
-            return mediaItem.buildUpon()
-                .setUri(mediaItem.mediaMetadata.mediaUri)
-                .build()
-        }
-    }
```
tzugen: Nice, this is not needed anymore?
It was a weird bit of architecture...

But then the comment can be removed as well

Maxmystere: From what I understood it is now directly done in onAddMediaItems(AutoMediaBrowserCallback.kt:184) but I'm not really sure it's necessary since it's also setup in MediaPlayerController.kt:662

Maxmystere:
Welp I'm stuck here, circleci is saying media3_notification_pause.xml is unused but it is used through media3

tzugen:
In this case you have a look if it's detekt or lint complaining. Probably lint. Then you remove the lint baseline file (which is like the list of allowed exceptions) and run lint locally again. Then lint will generate a new baseline file with the resource warning added

tzugen (approved):
Approved (once lint passes)

tzugen:
I added some info on how to do the lint stuff to the wiki:
https://github.com/ultrasonic/ultrasonic/blob/develop/CONTRIBUTING.md

Maxmystere:
Thanks for the wiki update !
It's here for now tho
https://github.com/ultrasonic/ultrasonic/blob/tzugen-patch-1/CONTRIBUTING.md

tzugen:
Thank you!

